PR #3, opened 2020-03-19
Gian populate counselors cors

GPaoloni:
Added CORS support.
Security method changed, now we are using JWT token auth vía twilio-flex-token-validator.


functions/populateCounselors.ts
```diff
@@ -0,0 +1,54 @@
...
+    response.appendHeader('Access-Control-Allow-Origin', '*');
+    response.appendHeader('Access-Control-Allow-Methods', 'OPTIONS, POST, GET');
+    response.appendHeader('Access-Control-Allow-Headers', 'Content-Type');
+
+    try {
+      const body = event as EventBody;
+      const { workspaceSID } = body;
+
+      if (workspaceSID === undefined) {
+        throw new Error('Error: Workspace parameter not provided');
```
nickhurlburt: How about we say "workspaceSID" to make it clear what parameter is needed?

functions/populateCounselors.ts
```diff
@@ -0,0 +1,54 @@
...
+      const prettyWorkers = workers.map(w => ({
+        sid: w.sid,
+        friendlyName: w.friendlyName,
+      }));
+
+      response.setStatusCode(200);
+      response.appendHeader('Content-Type', 'application/json');
+      response.setBody({ prettyWorkers });
+      callback(null, response);
+    } catch (err) {
```
nickhurlburt: This error handling needs to be better.  When I send a request without a `workspaceSID`, I get back a 500 and a body of `{}`.  We should get back a 400 (because the request is malformed) and get back the error message in the body.

In the case where the call to the Worker resource fails, we should return a 502 (because an upstream service failed) and _probably_ return the original error from the Worker resource in the body.

functions/populateCounselors.ts
```diff
@@ -0,0 +1,54 @@
...
+      response.setStatusCode(500);
+      response.appendHeader('Content-Type', 'application/json');
```
nickhurlburt: And this is setting the content type as JSON, but it doesn't look like our error message will come back as JSON as is.  Am I right?  Returning JSON is good, but our error message should actually be JSON.

PR #5, opened 2020-05-04
Localization via Twilio strings

GPaoloni:
This PR adds two functions (both protected by `twilio-flex-token-validator`):
`/getTranslation`: on POST, given a `language` parameter (in the body), returns the translations for the flex UI for the given language (if present).
`/getMessages`: on POST, given a `language` parameter (in the body), returns the translations for the system messages (used to send to the chat-contact) for the given language (if present).

Also adds some private assets (translations in json format) to fill this functions as examples for the localization process.


functions/populateCounselors.ts
```diff
@@ -8,34 +8,45 @@ import {
...
   callback(null, response);
 };
 
-type EventBody = {
+// TODO: Factor out into lib
+const responseWithCors = () => {
+  const response = new Twilio.Response();
+
+  response.setHeaders({
+    'Access-Control-Allow-Origin': '*',
```
nickhurlburt, line 25: Unlike the translations this does have info we don't want to openly share, and it'd be nice to start out with it a little more limited than `*`.

Unfortunately [it doesn't look like you can have a list of multiple domains here](https://stackoverflow.com/questions/1653308/access-control-allow-origin-multiple-origin-domains), but you could check the `origin` header in the request and return that domain.  So in bad pseudocode, something like:

```
const origin = context.request.getHeader('origin');
if (['http://localhost:3000', 'https://flex.twilio.com'].includes(origin) {
  response.setHeaders({
    'Access-Control-Allow-Origin': origin,
    ...
  });
} else {
  return no cors headers
};

Please take 15 minutes to see if that works.  If not, we can use `'*'` and create an issue to follow up later.  I realize that using the token validator does protect us, but multiple layers of security aren't a bad idea.

File: functions/populateCounselors.ts [hunk 2]
```diff
@@ -8,34 +8,45 @@ import {
 
 const TokenValidator = require('twilio-flex-token-validator').functionValidator;
 
-const send = (response: TwilioResponse) => (statusCode: number) => (body: string | object) => (
-  callback: ServerlessCallback,
+// TODO: Factor out into lib
+const send = (statusCode: number) => (body: string | object) => (callback: ServerlessCallback) => (
```
nickhurlburt, line 12: Is it better to curry all the parameters vs have a list of them?

assets/translations/es/messages.private.json
```diff
@@ -0,0 +1,64 @@
+{
```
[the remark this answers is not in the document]
nickhurlburt: Also just a note: when we get to providing translations for our current beta helplines (in Italian, Greek, Khmer, etc), I don't think all of these strings will need to be translated.  Just the strings that get exposed in the current chat app, which is only maybe 3 or 4 I think.

functions/populateCounselors.ts
GPaoloni: Here the idea of currying is because we can have something like `const error404 = send(404)({some error message})` and then use it like `error404(callback)(response);` (taking advantage of partial application). But I'll change it to a listed version if you prefer!

assets/translations/es/messages.private.json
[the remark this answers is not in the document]
GPaoloni: Yes, I though about this and dropped them in the same file, because this refers to what a caller will see (no matter the chat channel of choice). Should I split them in two separate files? Maybe translate the webchat strings within the code? Right now they don't share any string (webchat have html tags inside GoodByeMsg).

assets/translations/es/messages.private.json
```diff
@@ -0,0 +1,64 @@
...
+  "ConfirmEndEngagementMessage": "Finalizar este chat?",
+  "Connecting": "Conectando â€¦",
+  "DialpadAsteriskCaption": "*",
+  "DialpadAsteriskDescription": "",
+  "DialpadHashCaption": "#",
+  "DialpadHashDescription": "",
+  "DialpadNum0Caption": "0",
+  "DialpadNum0Description": "+",
+  "DialpadNum1Caption": "1",
+  "DialpadNum1Description": "â€”",
```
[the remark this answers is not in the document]
GPaoloni: Yeah, this was printed from the strings object and I left them because didn't knew if they were on purpose.. I'll remove those, but what should be here instead? It keeps printing the same character.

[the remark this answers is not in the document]
nickhurlburt: A quick google search turns up that [this is the long hyphen or em-dash](https://markmcb.com/2011/11/07/replacing-with-utf-8-characters-in-ruby-on-rails/).

functions/populateCounselors.ts
nickhurlburt: I see.  It's fine to have it this way, I just wanted to understand why.  Happy to keep as is given the description.

assets/translations/es/messages.private.json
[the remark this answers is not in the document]
nickhurlburt: Yeah, given that there's basically no overlap, I don't think they need to pull from the same file as the UI.  I'm okay with putting them right in the code here or in a separate file for now.  If we put them in the code they should be in the same kind of JSON format we would use in a file.

functions/getMessages.ts
```diff
@@ -0,0 +1,59 @@
...
+    'Access-Control-Allow-Methods': 'OPTIONS, POST, GET',
+    'Access-Control-Allow-Headers': 'Content-Type',
+    'Content-Type': 'application/json',
+  });
+
+  return response;
+};
+
+type Body = {
+  language: string | undefined;
```
murilovmachado, line 35: Can we use the `?` operator here: `language?: string`
(and other parts of the code with similar `| undefined` definitions)

GPaoloni: Yes. I was using the `| undefined` just to be more explicit, but if you are ok with the optional  operator, I'll change to that!

PR #8, opened 2020-05-27
Helpers refactor + added tests + setup CI

GPaoloni:
This PR factors out all the helpers, dropping them in the npm package called `tech-matters-serverless-helpers`, which code is included within this same repository (for ease), inside the folder `tech-matters-serverless-helpers-lib`. The name of the folder differs from the package cause we want `npm install tech-matters-serverless-helpers` to install the latest published package, not the local version.

Tests for functions, mocks and CI are also added.


.eslintignore
```diff
@@ -1,2 +1,2 @@
-functions/*.js
-utils/*.js
\ No newline at end of file
+**/*.js
```
nickhurlburt, line 1: Is this preventing ESLint from running on all files?

jest.config.js
```diff
@@ -0,0 +1,15 @@
...
+    },
+  },
+  moduleFileExtensions: ['ts', 'js'],
+  transform: {
+    '^.+\\.(ts|tsx)$': './node_modules/ts-jest',
+  },
+  testMatch: ['**/tests/**/*.test.(ts|js)'],
+  testEnvironment: 'node',
+  preset: 'ts-jest', // use this if you are using TypeScript
+  // globalTeardown: './jest.global-teardown.js' // optional: will be called once after all tests are executed
```
nickhurlburt, line 14: do we need to keep this commented-out code?

tests/setTaskInsightsData.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    await setTaskInsightsData(baseContext, event, callback);
+  });
+
+  test('Should return status 200 (with customers)', async () => {
+    const event: Body = {
+      workspaceSID: 'WSxxx',
+      taskSID: 'task1',
+      customers: '{"customer2":"customer2"}',
+      // conversations: '',
```
murilovmachado, line 114: I'd add the `conversations` object as well, with any `property: value`. Other than that, thanks for creating this tests, they look very good.

jest.config.js
GPaoloni: Not really, I'll remove it!

.eslintignore
murilovmachado: Yes, but I believe that's what we want in this project. We'll run ESLint on the `.ts` files.

GPaoloni: Yes, but in .js files. As this project is TS, we don't care on linting the JS (the code that TS compiler emits is readable but not linting compliant at all).

GPaoloni: Just what @murilovmachado said

PR #11, opened 2020-06-03
Add Sync Client for sharing contact form

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-142
This PR is related to https://github.com/tech-matters/flex-plugins/pull/106

Added two new functions:

- `issueSyncToken`: function that only expects the user to send a valid Twilio Token in the body, and returns a new token, with read/write access on the Sync Client that will be used for sharing state across counselors.
- `transferCallResolve`: given a `taskSid` and `reservationSid`, sets the state of the reservation to `completed`. Used to "close" the unnecessary task that stays in `wrapup` mode after a `WARM` transfer of a call task.


README.md
```diff
@@ -9,15 +9,27 @@ Repository for serverless functions living on the Twilio Serverless Toolkit
...
 For help on twilio-run commands run:  
 `npm run tr -- help`  
    
-To deploy:  
-`npm run tr:deploy`  
+## Environment variables
+| Variable Name                       | Expected Value                                |
+| ----------------------------------- | --------------------------------------------  |
+| `ACCOUNT_SID`                       | sid of the Twilio account                     |
+| `AUTH_TOKEN`                        | auth token of the abouve account              |
```
nickhurlburt, line 23: above

functions/issueSyncToken.ts
```diff
@@ -0,0 +1,66 @@
...
+const TokenValidator = require('twilio-flex-token-validator').functionValidator;
+
+type EnvVars = {
+  ACCOUNT_SID: string;
+  SYNC_SERVICE_API_KEY: string;
+  SYNC_SERVICE_API_SECRET: string;
+  SYNC_SERVICE_SID: string;
+};
+
+// This are added to event by TokenValidator (if a valid Token was provided) https://www.npmjs.com/package/twilio-flex-token-validator#token-result
```
nickhurlburt, line 18: This is

functions/transferCallResolve.ts
```diff
@@ -0,0 +1,66 @@
...
+type EnvVars = {
+  TWILIO_WORKSPACE_SID: string;
+};
+
+export type Body = {
+  taskSid?: string;
+  reservationSid?: string;
+};
+
+async function closeTask(context: Context<EnvVars>, body: Required<Body>) {
```
GPaoloni, line 26: I'm using this name because in the task is "gone" from the agent panel after this, but what is really happening here is that the reservation is being completed. Should I change the name for this function?

nickhurlburt: The function name does seem out of sync with the code inside of it, which has `closedReservation`.  So `closeReservation` seems to be more natural.

PR #12, opened 2020-06-10
Re-do the way transferred task are being closed

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-152
This PR is related to https://github.com/tech-matters/flex-plugins/pull/109

- Removed `transferCallResolve` function, as it is not required anymore
- Changed the chat transfers related functions. Instead of transitioning them to `completed`, they are set as `wrapping`, so TransferredTaskJanitor can clean them gracefully in Flex instance.


functions/transferChatStart.ts
```diff
@@ -48,7 +48,7 @@ async function closeTask(context: Context<EnvVars>, sid: string, taskToCloseAttr
     .workspaces(context.TWILIO_WORKSPACE_SID)
```
nickhurlburt, line 48: Should this variable still be named `closedTask`?

GPaoloni: I thought about this, and there is nothing else a counselor can do with the task after this, so that's why I preserved the name. But also would be great to change it to something that makes more sense! (same goes for the function `closeTask`)

PR #15, opened 2020-06-25
Capacity limit reached prevents chat transfers

GPaoloni:
Mimicking behavior of call transfers, when a chat transfer is attempted against a target worker who reached it's capacity limit (for the task channel of the task), the `transferChatStart` return error status 403, preventing the transfer.

### DIFF
--- a/functions/transferChatStart.ts
+++ b/functions/transferChatStart.ts
@@ -146,7 +146,7 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
         const workerChannel = await client.taskrouter
           .workspaces(context.TWILIO_WORKSPACE_SID)
           .workers(targetSid)
-          .workerChannels(originalTask.taskChannelSid)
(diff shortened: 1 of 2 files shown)


functions/transferChatStart.ts
```diff
@@ -136,13 +142,18 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
...
           return;
         }
-      }
 
-      // retrieve attributes of the original task
-      const originalTask = await client.taskrouter
-        .workspaces(context.TWILIO_WORKSPACE_SID)
-        .tasks(taskSid)
-        .fetch();
+        const workerChannel = await client.taskrouter
```
nickhurlburt, line 146: I wonder if we should make this call and the call on Line 136 in parallel to save time.  But that can be a future optimization.

GPaoloni: You are right, I'll change that before merging!

PR #19, opened 2020-07-23
Added blocklist code.

raarielgrace:


assets/blocklist.private.json
```diff
@@ -0,0 +1,3 @@
+{
+    "numbers": ["+12025550102"]
```
nickhurlburt, line 2: This is included in the PR, but is also in `.gitignore`.  Did you mean to `git rm` this?

PR #22, opened 2020-09-25
Added a function that pre-process chat tasks

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-249
This PR is related to https://github.com/tech-matters/webchat/pull/9

Adds a new attribute (`lockInput`) to the user resource. That attribute is used in webchat to hide the input (maybe the naming is wrong). The webchat is in charge of changing the value of `lockInput` once a counselor joins, using channel events.

Useful docs:
https://www.twilio.com/docs/chat/rest/user-resourcehttps://media.twiliocdn.com/sdk/js/chat/releases/3.2.3/docs/User.html#updateAttributes__anchor

### DIFF
--- a/functions/autopilotRedirect.protected.ts
+++ b/functions/autopilotRedirect.protected.ts
@@ -62,11 +62,16 @@ export const handler: ServerlessFunctionSignature<EnvVars, Event> = async (
   event: Event,
   callback: ServerlessCallback,
 ) => {
-  if (event.Channel === 'chat' && event.CurrentTask === 'redirect_function')
-    await handleChatChannel(context, event);
+  try {
+    if (event.Channel === 'chat' && event.CurrentTask === 'redirect_function')
+      await handleChatChannel(context, event);
 
-  const actions = buildActionsArray(context, event);
-  const returnObj = { actions };
+    const actions = buildActionsArray(context, event);
+    const returnObj = { actions };
 
-  callback(null, returnObj);
+    callback(null, returnObj);
+  } catch (err) {
+    // If something goes wrong, just handoff to counselor so contact is not lost
+    callback(null, { actions: [{ redirect: 'task://counselor_handoff' }] });
+  }
 };
--- a/functions/autopilotRedirect.protected.ts
+++ b/functions/autopilotRedirect.protected.ts
@@ -5,8 +5,9 @@ import {
 } from '@twilio-labs/serverless-runtime-types/types';
 
 export interface Event {
-  Memory: string;
   Channel: string;
+  CurrentTask: string;
+  Memory: string;
   UserIdentifier: string;
 }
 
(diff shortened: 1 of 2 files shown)


functions/autopilotRedirect.protected.ts
```diff
@@ -1,40 +1,95 @@
...
+  // if channel is webchat, disable the input
+  if (attributes.channel_type === 'web') {
+    const user = await context
+      .getTwilioClient()
+      .chat.services(ServiceSid)
+      .users(event.UserIdentifier)
+      .fetch();
+
+    const userAttr = JSON.parse(user.attributes);
+    const updatedAttr = { ...userAttr, lockInput: true };
```
nickhurlburt, line 38: This seems to lock the input in every situation, even if input is needed for the chatbot.  Currently on staging chat, you get stuck because it asks a question but the input is locked.  Is that an issue with this?

image

GPaoloni: That is the behavior of webchat with the actual version of staging: the input is hidden as soon as the user starts the chat (regardless of the presence of the chatbot), and it's re-enabled after a counselor joins the chat (which will never happen as we are stuck with the bot). To test the functionality in this PR, you must use the chat version in https://github.com/tech-matters/webchat/pull/9

functions/autopilotRedirect.protected.ts
```diff
@@ -1,40 +1,98 @@
...
+            uri: `taskrouter://${context.TWILIO_CHAT_TRANSFER_WORKFLOW_SID}`,
+          },
+        };
+        return [say, handoff];
+      }
+
+      const redirect = { redirect: 'task://counselor_handoff' };
+      return [redirect];
+    }
+    case 'gender_why': {
```
nickhurlburt, line 64: Turns out that 'gender_why' is no longer used, so this can be removed.

functions/autopilotRedirect.protected.ts
```diff
@@ -1,40 +1,98 @@
...
+        const say = { say: 'Hold on, we are connecting you with an agent' };
+        const handoff = {
+          handoff: {
+            channel: 'voice',
+            uri: `taskrouter://${context.TWILIO_CHAT_TRANSFER_WORKFLOW_SID}`,
+          },
+        };
+        return [say, handoff];
+      }
+
```
nickhurlburt, line 60: Since we are not doing voice for now, I think we should remove this.  But that is a suggestion only -- I'm open to keeping it.

nickhurlburt (approved):
Looks good!  Once this is out, can you put it on production-staging and update autopilot/studio so that we don't forget?

GPaoloni:
Deployed to staging account, prod environment (chatbot updated to make use of it).

PR #23, opened 2020-09-30
Added the function that handles capacity change for manual pull

nickhurlburt:
Supports https://github.com/tech-matters/flex-plugins/pull/212 by implementing increment and decrement of current capacity.


nickhurlburt:
Subsumed by #24.

PR #25, opened 2020-10-07
Continue to support transfers with manual pull

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-256
This PR is related to: https://github.com/tech-matters/flex-plugins/pull/219

- Moved chat channel finding logic to the `adjustChatCapacity` function.
- Added integration for manual pull with transfers: if a counselor has availavable `worker.attributes.maxMessageCapacity`, the chat channel capacity is increased from within `transferChatStart`.

### DIFF
--- a/functions/adjustChatCapacity.ts
+++ b/functions/adjustChatCapacity.ts
@@ -34,6 +34,8 @@ export const adjustChatCapacity = async (
     .workers(body.workerSid)
     .fetch();
 
+  if (!worker) return { status: 404, message: 'Could not find worker.' };
+
   const { maxMessageCapacity } = JSON.parse(worker.attributes);
 
   const channels = await worker.workerChannels().list();
--- a/functions/adjustChatCapacity.ts
+++ b/functions/adjustChatCapacity.ts
@@ -14,33 +14,38 @@ import {
 
 const TokenValidator = require('twilio-flex-token-validator').functionValidator;
 
-type EnvVars = {};
+type EnvVars = {
+  TWILIO_WORKSPACE_SID: string;
+};
 
 export type Body = {
   workerSid?: string;
-  workspaceSid?: string;
-  channelSid?: string;
-  workerLimit?: number;
   adjustment?: 'increase' | 'decrease';
 };
 
-const adjustChatCapacity = async (
+export const adjustChatCapacity = async (
   context: Context<EnvVars>,
   body: Required<Body>,
 ): Promise<{ status: number; message: string }> => {
   const client = context.getTwilioClient();
 
-  const channel = await client.taskrouter
-    .workspaces(body.workspaceSid)
+  const worker = await client.taskrouter
+    .workspaces(context.TWILIO_WORKSPACE_SID)
     .workers(body.workerSid)
-    .workerChannels(body.channelSid)
     .fetch();
+
+  const { maxMessageCapacity } = JSON.parse(worker.attributes);
+
+  const channels = await worker.workerChannels().list();
+  const channel = channels.find(c => c.taskChannelUniqueName === 'chat');
+
+  if (!channel) return { status: 404, message: 'Could not find chat channel.' };
 
   if (body.adjustment === 'increase') {
     if (channel.availableCapacityPercentage > 0)
       return { status: 412, message: 'Still have available capacity, no need to increase.' };
 
-    if (!(channel.configuredCapacity < body.workerLimit))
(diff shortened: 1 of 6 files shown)


functions/adjustChatCapacity.ts
```diff
@@ -14,33 +14,40 @@ import {
...
 
-  const channel = await client.taskrouter
-    .workspaces(body.workspaceSid)
+  const worker = await client.taskrouter
+    .workspaces(context.TWILIO_WORKSPACE_SID)
     .workers(body.workerSid)
-    .workerChannels(body.channelSid)
     .fetch();
 
+  if (!worker) return { status: 404, message: 'Could not find worker.' };
```
nickhurlburt, line 37: Will the error here and at L44 be logged in Twilio debugger or in Rollbar (via the Flex plugin) so that we will know about them?  By moving the worker and channel lookup to this function, we are creating more remote API calls, and more things that can fail.  That is okay if we have a way to detect when that happens.

functions/adjustChatCapacity.ts
```diff
@@ -14,33 +14,40 @@ import {
...
+
+  const { maxMessageCapacity } = JSON.parse(worker.attributes);
```
nickhurlburt, line 39: What happens if this value is missing?

functions/transferChatStart.ts
```diff
@@ -93,6 +100,81 @@ async function closeTaskAndKick(context: Context<EnvVars>, body: Required<Body>)
...
+      payload: { status: 403, message: "Error: can't transfer to an offline counselor" },
+    } as const;
+
+  const workerAttr = JSON.parse(worker.attributes);
+
+  const unavailableVoice = channelType === 'voice' && !workerChannel.availableCapacityPercentage;
+  const unavailableChat =
+    channelType !== 'voice' &&
+    !workerChannel.availableCapacityPercentage &&
+    workerAttr.maxMessageCapacity &&
```
nickhurlburt, line 139: I think this function needs a test for when `maxMessageCapacity` has not been set.  It is possible that whoever is configuring the system could overlook setting it for a particular worker.  I think a reasonable default in that case should be to rely only on the configured capacity as is.  But here I think `unavailableChat` will be `false` if `maxMessageCapacity` is `null`, which would be misleading.

nickhurlburt (commented):
This looks good, though there's some further need to consider error handling and failure cases.

functions/adjustChatCapacity.ts
GPaoloni: Yes. I intentionally triggered that (setting the condition to true). When we manage to connect Rollbar to debugger hook, this will be fired as an event, and it's still in Twilio's console in the meanwhile.
Screenshot from 2020-10-07 16-32-25

functions/adjustChatCapacity.ts
GPaoloni: Addressed in https://github.com/tech-matters/serverless/pull/25/commits/7eefe7365dd3bb35ac6e29f0c1839c377ae8421a

functions/transferChatStart.ts
GPaoloni: Addressed in https://github.com/tech-matters/serverless/pull/25/commits/b3ede0c809d8df1b0c04a8aa94830b1941bae5ef

functions/transferChatStart.ts
```diff
@@ -133,11 +133,14 @@ async function validateChannelIfWorker(
   const workerAttr = JSON.parse(worker.attributes);
 
   const unavailableVoice = channelType === 'voice' && !workerChannel.availableCapacityPercentage;
+  // if maxMessageCapacity is not set, just use configuredCapacity without adjustChatCapacity
   const unavailableChat =
-    channelType !== 'voice' &&
+    workerAttr.maxMessageCapacity 
+      ? channelType !== 'voice' &&
     !workerChannel.availableCapacityPercentage &&
     workerAttr.maxMessageCapacity &&
```
nickhurlburt, line 141: This line is now redundant (since `workerAttr.maxMessageCapacity` must be truthy if you got here)

nickhurlburt (approved):
Looks good!  There's one redundant line that can be removed, but that's it.

functions/transferChatStart.ts
GPaoloni: Thanks for the catch!

GPaoloni:
Deployed to staging.

nickhurlburt:
Yay!

PR #27, opened 2020-11-06
Added createContactlessTask function

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-305 + https://bugs.benetech.org/browse/CHI-311
This PR is related to https://github.com/tech-matters/flex-plugins/pull/242

This task adds the function createContactlessTask, that creates a new task withouth contact (no chat nor voice capabilities). This is achieved by using the `default` channel that Twilio provides, and is not taken into account when dealing with chat capacities.

Note: this PR includes auto-generated changes in swagger file.


functions/createContactlessTask.ts
```diff
@@ -0,0 +1,74 @@
...
+        return;
+      }
+      if (helpline === undefined) {
+        resolve(error400('helpline'));
+        return;
+      }
+
+      const newAttributes = {
+        targetSid,
+        transferTargetType,
```
nickhurlburt, line 53: Ah... this is reusing our transfer functionality.  Great!

nickhurlburt (approved):
Looks good.  Are there any edge cases to consider?
What if the worker is unavailable?
Do we need to ensure that the `default` capacity is set to one?
What if the worker rejects the task?
Anything else?

GPaoloni:
> What if the worker is unavailable?

Right now, the task is created anyways. Should we prevent this?

> Do we need to ensure that the default capacity is set to one?

Actually I don't think so, the frontend is not letting you create a new task if there is already one, but we can limit that from worker capacity too, if desired.

> What if the worker rejects the task?

Right now that's not being considered. If the worker rejects the task it will be automatically reassigned as is the only worker that matches. It can be canceled after accepting the reservation.

nickhurlburt:
> > What if the worker is unavailable?

> Right now, the task is created anyways. Should we prevent this?

I think we should block it either here or in the plugin.

> > Do we need to ensure that the default capacity is set to one?

> Actually I don't think so, the frontend is not letting you create a new task if there is already one, but we can limit that from worker capacity too, if desired.

If it's prevented in the frontend, that should be fine.

> > What if the worker rejects the task?

> Right now that's not being considered. If the worker rejects the task it will be automatically reassigned as is the only worker that matches. It can be canceled after accepting the reservation.

That sounds fine.  Just want to be aware of what will happen.

GPaoloni:
Deployed to staging.

PR #28, opened 2021-01-29
Security updates

nickhurlburt:
@GPaoloni this is just from running `npm audit fix`.

After you've taken a quick look, could you deploy to staging?  The internal npm package may also need an update.


GPaoloni (approved):

GPaoloni:
Deployed to staging.

PR #29, opened 2021-02-03
Make the "Counselor has left the chat" message consistent across all text-based channels

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-314
This PR is related to
- https://github.com/tech-matters/webchat/pull/15
- https://github.com/tech-matters/flex-plugins/pull/349

This PR adds `sendSystemMessage` function, which takes in `taskSid`, `message` and (optionally) `from` parameters, and send a `message` to the channel bounded to the given `taskSid` in behalf of `from`. Default from is "system".

### DIFF
--- a/functions/sendSystemMessage.ts
+++ b/functions/sendSystemMessage.ts
@@ -1,0 +1,66 @@
+import '@twilio-labs/serverless-runtime-types';
+import {
+  Context,
+  ServerlessCallback,
+  ServerlessFunctionSignature,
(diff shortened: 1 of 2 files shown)


andvirga (approved):
I don't have too much experience on this. But the chat behavior is the expected! Great work!

nickhurlburt (approved):
Looks good!  I am surprised that we don't need to add a member to the channel before sending a message.

PR #30, opened 2021-02-10
Create south_africa_staging.yml

fgonzalez-techmatters:
Add GitHub action to deploy Serverless to South Africa staging environment.

### DIFF
--- a/.github/workflows/south_africa_staging.yml
+++ b/.github/workflows/south_africa_staging.yml
@@ -24,11 +24,17 @@ jobs:
       # Runs a single command using the runners shell
       - name: Install dependencies for the twilio function
         run: npm ci
-      - name: Install twilio cli and run deploy command
+      # Compile typescript to javascript
+      - name: Transpile typescript to javascript
+        run: npx tsc
+      # Deploy the function        
+      - name: Run deploy command to dev
         env:
          ACCOUNT_SID: ${{ secrets.SA_STG_ACCOUNT_SID }}
          AUTH_TOKEN: ${{ secrets.SA_STG_AUTH_TOKEN }}
-        # Install the Twilio CLI and the serverless plugin then deploy the function
-        run: npm run tr -- deploy --service-name=serverless --environment=dev --force
-      - name: deploy to Production
-        run: npm run tr -- deploy --service-name=serverless --environment=production --force
+        run: npm run tr -- deploy --service-name=serverless --environment=dev --account-sid $ACCOUNT_SID --auth-token $AUTH_TOKEN --force
+      - name: Run deploy command to Production
+        env:
+         ACCOUNT_SID: ${{ secrets.SA_STG_ACCOUNT_SID }}
+         AUTH_TOKEN: ${{ secrets.SA_STG_AUTH_TOKEN }}      
+        run: npm run tr -- deploy --service-name=serverless --environment=production --account-sid $ACCOUNT_SID --auth-token $AUTH_TOKEN --force
--- a/.github/workflows/south_africa_staging.yml
+++ b/.github/workflows/south_africa_staging.yml
@@ -1,0 +1,34 @@
+# This is a basic workflow to publish a Twilio function with Github Actions
+
+name: Serverless release to South Africa Staging.
+
+# Controls when the action will run.
+on:
+  workflow_dispatch
+  
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


.github/workflows/south_africa_staging.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      # Runs a single command using the runners shell
+      - name: Install dependencies for the twilio function
+        run: npm ci
+      - name: Install twilio cli and run deploy command
+        env:
+         TWILIO_ACCOUNT_SID: ${{ secrets.SA_STG_ACCOUNT_SID }}
+         TWILIO_AUTH_TOKEN: ${{ secrets.SA_STG_AUTH_TOKEN }}
```
murilovmachado, line 30: Shouldn't the env variables be:
`ACCOUNT_SID` instead of `TWILIO_ACCOUNT_SID`
`AUTH_TOKEN` instead of `TWILIO_AUTH_TOKEN`

That's how it is in the docs, but it may be outdated.

.github/workflows/south_africa_staging.yml
```diff
@@ -0,0 +1,34 @@
...
+        # Install the Twilio CLI and the serverless plugin then deploy the function
+        run: npm install twilio-cli -g && twilio plugins:install @twilio-labs/plugin-serverless && twilio serverless:deploy --service-name=serverless --environment=dev --force
```
murilovmachado, line 32: Could this line be just this:
`npm run tr -- deploy --service-name=serverless --environment=dev --force`

Behind the scenes, this line is calling `twilio-run`, that's already installed on line 26.
This way, we wouldn't need to install `twilio-cli` nor `@twilio-labs/plugin-serverless`.

.github/workflows/south_africa_staging.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: deploy to Production
+        run: twilio serverless:deploy --service-name=serverless --environment=production --force
```
murilovmachado, line 34: and this line could be almost the same as the example above:
`npm run tr -- deploy --service-name=serverless --environment= production --force`

murilovmachado (changes requested):
Overral it looks good. But I have two questions below:
1) If the env var names are correct (this is the most important thing)
2) Reuse what we have in package.json, so we don't need to install extra stuff for deploying.

It would be good to consult @GPaoloni on this PR since he has worked more on the serverless repo.

.github/workflows/south_africa_staging.yml
GPaoloni: Just checked and I confirm, we are using `ACCOUNT_SID` and `AUTH_TOKEN` like stated in the docs.

.github/workflows/south_africa_staging.yml
GPaoloni: Seems to me that what @murilovmachado says here makes sense.

GPaoloni (commented):
I agree with the comments @murilovmachado left here. One important thing to add is that this repo is TS, but only JS files are deployed to the functions environment, so you need to run `tsc` before the deployment takes place.
That's why we first run this script before deploying to production env.
Screenshot from 2021-02-10 14-45-12

.github/workflows/south_africa_staging.yml
fgonzalez-techmatters: Thanks for checking this. I will correct it to use the right format.

.github/workflows/south_africa_staging.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install dependencies for the twilio function
+        run: npm ci
+      - name: Install twilio cli and run deploy command
+        env:
+         ACCOUNT_SID: ${{ secrets.SA_STG_ACCOUNT_SID }}
+         AUTH_TOKEN: ${{ secrets.SA_STG_AUTH_TOKEN }}
+        # Install the Twilio CLI and the serverless plugin then deploy the function
+        run: npm run tr -- deploy --service-name=serverless --environment=dev --force
+      - name: deploy to Production
+        run: npm run tr -- deploy --service-name=serverless --environment=production --force
```
murilovmachado, line 34: Looks good! I think there's only one last thing missing, that is compiling typescript to javascript, as @GPaoloni said.

After the install dependencies step, you could create another step like:
```
- name: Transpile typescript to javascript
   run: tsc
```
Maybe you'll need to `npm i typescript` in order for `tsc` to work.
My suggestion would be to add a script to package.json:
`"compile": "tsc"`
and than you can have the following step in the yaml file:
```
- name: Transpile typescript to javascript
   run: npm run compile
```

murilovmachado (approved):

murilovmachado (approved):

PR #31, opened 2021-02-11
Add commented step for newer serverless versions. Add workflows for a…

fgonzalez-techmatters:
…ll accounts

Added a GitHub action per Twilio account, to deploy Serverless repo. Added a commented step in every action with the commands to deploy newer versions of the Serverless function, to be able to switch it in the future.


GPaoloni (approved):
Looks great to me!

PR #32, opened 2021-02-19
Add assignOfflineContact function

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-500
This PR is related to https://github.com/tech-matters/flex-plugins/pull/361

This PR creates a new task in behalf of a target worker (counselor), in order to be consistent between HRM and Insights reports. 
- If the worker is offline, it's attributes are changed to apply a new routing rule that will prevent other tasks to be reserved (refer to Admin workflow in Development environment to see the complete assignment expression):
 `OR (worker.waitingOfflineContact == true AND worker.acceptOnlyTask == task.sid)`
- Even if the counselor is currently active and reached the assigned tasks limit, we have no need to increase channel capacity as we are using "default" channel for this kind of tasks.

Resolved questions:
- Q: Should a counselor be able to issue a contact in behalf of another counselor when the target is online?
  A: Yes.

~WIP~:
- ~Handle rules to avoid other tasks to be reserved in the short interval the target counselor is set to available.~
- ~Test~

### DIFF
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -10,7 +10,10 @@ import {
   error400,
   error500,
   success,
+  send,
 } from '@tech-matters/serverless-helpers';
+// eslint-disable-next-line prettier/prettier
+import type { PromiseValue } from 'type-fest';
 
 const TokenValidator = require('twilio-flex-token-validator').functionValidator;
 
@@ -22,6 +25,57 @@ type EnvVars = {
 export type Body = {
   targetSid?: string;
   finalTaskAttributes: string;
+};
+
+// eslint-disable-next-line prettier/prettier
+type TaskInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['tasks']>['fetch']>>;
+// eslint-disable-next-line prettier/prettier
+type WorkerInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['workers']>['fetch']>>;
+
+const assignToAvailableWorker = async (
+  event: Body,
+  newTask: TaskInstance,
+) => {
+  const reservations = await newTask.reservations().list();
+  const reservation = reservations.find(r => r.workerSid === event.targetSid);
+
+  if (reservation) {
+    await reservation.update({ reservationStatus: 'accepted' });
+    await reservation.update({ reservationStatus: 'completed' });
+    return { type: 'success' } as const;
+  } 
+
+  await newTask.remove();
+  return {
+    type: 'error',
+    payload: { status: 500, message: 'Error: reservation for task not created.' },
+  } as const;
+  
+};
+
+const assignToOfflineWorker = async (
+  context: Context<EnvVars>,
+  event: Body,
+  targetWorker: WorkerInstance,
+  newTask: TaskInstance,
+) => {
+  const previousActivity = targetWorker.activitySid;
+  const previousAttributes = JSON.parse(targetWorker.attributes);
+  const availableActivity = await context
+    .getTwilioClient()
+    .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
+    .activities.list({ friendlyName: 'Available' });
+
+  await targetWorker.update({
+    activitySid: availableActivity[0].sid,
+    attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true, acceptOnlyTask: newTask.sid }), // waitingOfflineContact & acceptOnlyTask are routing rules used to avoid other tasks to be assigned during this window of time
+  });
+
+  const result = await assignToAvailableWorker(event, newTask);
+
+  await targetWorker.update({ activitySid: previousActivity, attributes: JSON.stringify(previousAttributes) });
+
+  return result;
 };
 
 export const handler: ServerlessFunctionSignature = TokenValidator(
@@ -59,27 +113,24 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
           priority: 100,
         });
 
-      const targeWorker = await client.taskrouter
+      const targetWorker = await client.taskrouter
         .workspaces(context.TWILIO_WORKSPACE_SID)
         .workers(targetSid)
         .fetch();
 
-      // Set the worker available, assign the task, accept & complete it and set worker to previous state
-      if (!targeWorker.available) {
-        const previousActivity = targeWorker.activitySid;
-        const availableActivity = await client.taskrouter
-          .workspaces(context.TWILIO_WORKSPACE_SID)
-          .activities.list({ friendlyName: 'Available' });
-        await targeWorker.update({ activitySid: availableActivity[0].sid });
+      let assignmentResult: PromiseValue<ReturnType<typeof assignToAvailableWorker>>;
+      if (targetWorker.available) {
+        // assign the task, accept and complete it
+        assignmentResult = await assignToAvailableWorker(event, newTask);
+      } else {
+        // Set the worker available, assign the task, accept, complete it and set worker to previous state
+        assignmentResult = await assignToOfflineWorker(context, event, targetWorker, newTask);
+      }
 
-        const reservations = await newTask.reservations().list();
-        const reservation = reservations.find(r => r.workerSid === targetSid);
-        if (reservation) {
-          await reservation.update({ reservationStatus: 'accepted' });
-          await reservation.update({ reservationStatus: 'completed' });
-        }
-
-        await targeWorker.update({ activitySid: previousActivity });
+      if (assignmentResult.type === 'error') {
+        const { status, message } = assignmentResult.payload;
+        resolve(send(status)({ status, message }));
+        return;
       }
 
       resolve(success(newTask));
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -73,9 +73,10 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
         await targeWorker.update({ activitySid: availableActivity[0].sid });
 
         const reservations = await newTask.reservations().list();
-        if (reservations.length && reservations[0].workerSid === targetSid) {
-          await reservations[0].update({ reservationStatus: 'accepted' });
-          await reservations[0].update({ reservationStatus: 'completed' });
+        const reservation = reservations.find(r => r.workerSid === targetSid);
+        if (reservation) {
+          await reservation.update({ reservationStatus: 'accepted' });
+          await reservation.update({ reservationStatus: 'completed' });
         }
 
         await targeWorker.update({ activitySid: previousActivity });
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -21,8 +21,7 @@ type EnvVars = {
 
 export type Body = {
   targetSid?: string;
-  transferTargetType?: string;
-  helpline?: string;
+  finalTaskAttributes: string;
 };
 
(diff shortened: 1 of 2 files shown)


nickhurlburt:
@dee-luo what are your thoughts on the above question?

`Should a counselor be able to issue a contact in behalf of another counselor when the target is online?`

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,89 @@
...
+      const newTask = await client.taskrouter
+        .workspaces(context.TWILIO_WORKSPACE_SID)
+        .tasks.create({
+          workflowSid: context.TWILIO_CHAT_TRANSFER_WORKFLOW_SID,
+          taskChannel: 'default',
+          attributes: JSON.stringify(newAttributes),
+          priority: 100,
+        });
+
+      const targeWorker = await client.taskrouter
```
nickhurlburt, line 62: spelling: `targetWorker`

dee-luo:
> @dee-luo what are your thoughts on the above question?
> 
> `Should a counselor be able to issue a contact in behalf of another counselor when the target is online?`

Is the target the counselor being added to the offline contact? If so, then yes, it shouldn't matter what the target counselor's status is.

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,89 @@
...
+        .workers(targetSid)
+        .fetch();
+
+      // Set the worker available, assign the task, accept & complete it and set worker to previous state
+      if (!targeWorker.available) {
+        const previousActivity = targeWorker.activitySid;
+        const availableActivity = await client.taskrouter
+          .workspaces(context.TWILIO_WORKSPACE_SID)
+          .activities.list({ friendlyName: 'Available' });
+        await targeWorker.update({ activitySid: availableActivity[0].sid });
```
nickhurlburt, line 73: Here and in L81, it is worth checking that `availableActivity[0].sid !== previousActivity` to reduce unnecessary work?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,89 @@
...
+          priority: 100,
+        });
+
+      const targeWorker = await client.taskrouter
+        .workspaces(context.TWILIO_WORKSPACE_SID)
+        .workers(targetSid)
+        .fetch();
+
+      // Set the worker available, assign the task, accept & complete it and set worker to previous state
+      if (!targeWorker.available) {
```
nickhurlburt, line 68: I don't know if it will really be practical to expect the helpline to keep this constraint.

nickhurlburt (commented):
This looks good as a start.  Of course as you mention, we need to check for several edge cases.  Especially ensuring that more tasks don't get assigned for the brief moment the worker is available.  And I suspect that we can't count on the worker being unavailable... even if they are not on shift they may not have closed out.

We'll also need some tests... I am not sure how practical it will be to write code to test all the edge cases.

functions/assignOfflineContact.ts
murilovmachado, line 68: In case the `targetWorker` is already `available`, nothing would happen?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,89 @@
...
+  async (context: Context<EnvVars>, event: Body, callback: ServerlessCallback) => {
+    const client = context.getTwilioClient();
+
+    const response = responseWithCors();
+    const resolve = bindResolve(callback)(response);
+
+    const { targetSid, finalTaskAttributes } = event;
+
+    try {
+      if (targetSid === undefined) {
```
andvirga, line 37: Not big deal, but you can use: 
`if (!targetSid) { ... } `

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,89 @@
...
+        resolve(error400('targetSid'));
+        return;
+      }
+      if (finalTaskAttributes === undefined) {
```
andvirga, line 41: Same as above

functions/assignOfflineContact.ts
GPaoloni: There are cases where we want to be able to send empty strings, and checking for `if (!targetSid) { ... }` won't allow it. That's why, in general, in all the Serverless functions you will see me checking like this at first, then if the empty string (or whatever) is not valid, that error can be addressed in more detail. This just checks that the parameter at least exists.

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export type Body = {
+  targetSid?: string;
+  finalTaskAttributes: string;
+};
+
+// eslint-disable-next-line prettier/prettier
+type TaskInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['tasks']>['fetch']>>;
+// eslint-disable-next-line prettier/prettier
+type WorkerInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['workers']>['fetch']>>;
```
GPaoloni, line 33: Sorry this ugly types, couldn't pull them from Twilio libs and found this the easiest way to use them.

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+    .getTwilioClient()
+    .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
+    .activities.list({ friendlyName: 'Available' });
+
+  await targetWorker.update({
+    activitySid: availableActivity[0].sid,
+    attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true, acceptOnlyTask: newTask.sid }), // waitingOfflineContact & acceptOnlyTask are routing rules used to avoid other tasks to be assigned during this window of time
+  });
+
+  const result = await assignToAvailableWorker(event, newTask);
```
murilovmachado, line 74: Nice reusing.

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+        .fetch();
+
+      let assignmentResult: PromiseValue<ReturnType<typeof assignToAvailableWorker>>;
+      if (targetWorker.available) {
+        // assign the task, accept and complete it
+        assignmentResult = await assignToAvailableWorker(event, newTask);
+      } else {
+        // Set the worker available, assign the task, accept, complete it and set worker to previous state
+        assignmentResult = await assignToOfflineWorker(context, event, targetWorker, newTask);
+      }
```
murilovmachado, line 128: Readability is much better now!

murilovmachado (approved):
Code looks good to me, Gian. But I'm not sure what's the best way to test it.

andvirga (approved):
Great job! The code looks very good!

nickhurlburt:
Have we been keeping up with Swagger?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+type TaskInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['tasks']>['fetch']>>;
+// eslint-disable-next-line prettier/prettier
+type WorkerInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['workers']>['fetch']>>;
+
+const assignToAvailableWorker = async (
+  event: Body,
+  newTask: TaskInstance,
+) => {
+  const reservations = await newTask.reservations().list();
+  const reservation = reservations.find(r => r.workerSid === event.targetSid);
```
nickhurlburt, line 40: What if the worker got a real transfer at the same time?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  if (reservation) {
+    await reservation.update({ reservationStatus: 'accepted' });
```
nickhurlburt, line 43: Is it possible that the worker could see this come up in the All Tasks list and accept it before this code runs?  What would happen?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+    await reservation.update({ reservationStatus: 'completed' });
+    return { type: 'success' } as const;
+  } 
+
+  await newTask.remove();
+  return {
+    type: 'error',
+    payload: { status: 500, message: 'Error: reservation for task not created.' },
```
nickhurlburt, line 51: How will we handle this error when it is sent back to the frontend of the worker who entered the offline contact?

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const assignToOfflineWorker = async (
+  context: Context<EnvVars>,
+  event: Body,
+  targetWorker: WorkerInstance,
+  newTask: TaskInstance,
+) => {
+  const previousActivity = targetWorker.activitySid;
+  const previousAttributes = JSON.parse(targetWorker.attributes);
+  const availableActivity = await context
```
nickhurlburt, line 64: This may benefit from a type declaration, since I was surprised to discover that it's an array on line 70

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+    .getTwilioClient()
+    .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
+    .activities.list({ friendlyName: 'Available' });
+
+  await targetWorker.update({
+    activitySid: availableActivity[0].sid,
```
nickhurlburt, line 70: Do we need to guard against `availableActivity[0]` being `undefined`, and thus this will error?  Or maybe that's just being too paranoid.

functions/assignOfflineContact.ts
```diff
@@ -0,0 +1,141 @@
...
+    attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true, acceptOnlyTask: newTask.sid }), // waitingOfflineContact & acceptOnlyTask are routing rules used to avoid other tasks to be assigned during this window of time
```
nickhurlburt, line 71: > waitingOfflineContact & acceptOnlyTask are routing rules used to avoid other tasks

Do we need to add any TaskRouter rules to our documentation for setting up new environments?

nickhurlburt (approved):
Several questions, but overall looks great!

functions/assignOfflineContact.ts
GPaoloni: This are the reservation for the newly created task only, so if there are other tasks around, this should not interfere. Also given that we are using default channel, this kind of task will skip the "waiting queue" even if the counselor is at max capacity.

functions/assignOfflineContact.ts
GPaoloni: This takes a few miliseconds, I don't think that's possible, but we maybe we can "disable the UI" for that case.

functions/assignOfflineContact.ts
GPaoloni: The error 500 is handled in the frontend in the same way as the error in hrm, "backend error" pop-up alert. You can try it out by running the frontend with a bad serverless url.

functions/assignOfflineContact.ts
GPaoloni: Yes! The additional expression is in the description of this PR, so we don't lose it. I've added it in dev environment only so far.

functions/assignOfflineContact.ts
nickhurlburt: Oh right!  I missed that this is just reservations on the `newTask`

functions/assignOfflineContact.ts
nickhurlburt: Okay -- we'll have to make sure we remember to do that when deploying this to the different accounts

PR #35, opened 2021-03-23
Run on Node 12

nickhurlburt:
Twilio requires that we update all our functions to run on Node 12 by the end of April.  This PR does the following:
- Changes our deployment actions to use the Twilio CLI and sets a flag to run with Node 12
- Updates `package.json` to use the Node 12 engine
- Sets our CI to run with Node 12

Once this is merged, the next deploy to any environment will deploy to run on Node 12.  This is sufficient for Twilio's requirements.

Primary reviewer: @GPaoloni 
Also good to hear from @fgonzalez-techmatters


GPaoloni:
Per [npm docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines), I believe we should also update this statement in `package.json`:
Screenshot from 2021-03-24 13-56-58

Also is worth having [this doc](https://www.twilio.com/docs/runtime/twilio-functions-node-v12-upgrade#using-functions-ui-editor) in mind.

GPaoloni (approved):
We'll know if the deployment changes works only after trying a deployment, but looks great!

fgonzalez-techmatters (approved):

PR #36, opened 2021-03-30
Better handling in assignOfflineContact for stuck tasks

GPaoloni:
Note: reordered some code. Maybe is easier to read the entire file (starting in the handler implementation, at bottom of the file) rather than the diff in GH.

This PR adds more handling to stuck statuses for the tasks being created:
- No task is created (and returns error status code) if worker is already waiting for an offline contact (prevent simultaneous offline contacts. This can be removed if not desired tho).
- If the task is created but the reservation is not, or the reservation is created but not accepted, or is accepted but not completed, we clean out the newly created task and send the attributes to the issuer (debug).
- If the worker was set to available for this process only, rejects all possible pending reservations before going back to offline mode.

WIP: 
- CI (should I address this?)

### DIFF
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -96,7 +96,7 @@ type AssignmentResult = {
   payload: { status: number, message: string, taskRemoved: boolean, attributes?: string }
 } |  { type: 'success', newTask: TaskInstance };
 
-const assingOfflineContact = async (context: Context<EnvVars>, body: Required<Body>): Promise<AssignmentResult> => {
+const assignOfflineContact = async (context: Context<EnvVars>, body: Required<Body>): Promise<AssignmentResult> => {
   const client = context.getTwilioClient();
   const { targetSid, finalTaskAttributes } = body;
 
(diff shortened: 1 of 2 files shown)


functions/assignOfflineContact.ts
```diff
@@ -61,33 +73,78 @@ const assignToOfflineWorker = async (
...
+
   const availableActivity = await context
     .getTwilioClient()
     .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
     .activities.list({ friendlyName: 'Available' });
 
   await targetWorker.update({
     activitySid: availableActivity[0].sid,
-    attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true, acceptOnlyTask: newTask.sid }), // waitingOfflineContact & acceptOnlyTask are routing rules used to avoid other tasks to be assigned during this window of time
+    attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true }), // waitingOfflineContact is used to avoid other tasks to be assigned during this window of time (workflow rules)
```
nickhurlburt, line 84: I'm not sure if we should check for error here and on line 89 as well, but happy to go without for now.

nickhurlburt (approved):
Looks good

PR #39, opened 2021-04-07
Create customizable function to handle operating hours

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-532

This PR adds customizable `operatingHours` function. The customization can be done through a json file under the assets folder. We can then have all of the operating hours bundled within our assets, refer to the appropriate one via envvar and update them with a deploy, but that requires no code changes. 

To test this PR:
- Add `OPERATING_INFO_KEY=za` to your local `.env` file.
- For ease, you can set `const TokenValidator = (f: any) => f;` or remove the `TokenValidator` wrapper (that way you don't need to provide a token). If you do this you also must remove the `: ServerlessFunctionSignature` type annotation in `export const handler: ServerlessFunctionSignature`.
- Start the local development server with `npm run start`.
- Hit `operatingHours` and see response (`curl -X GET 'http://localhost:3000/operatingHours'`).
- You can check the current time in ZA [here](https://www.google.com/search?q=africa%2Fjohannesburg+timezone&sxsrf=ALeKk03m0x_gH9vAgYeWMGwmgddzIc-s-A%3A1617825791066&ei=_w9uYJvDA8XR5OUPmeiH4AM&oq=africa%2Fjohannesburg+timezo&gs_lcp=Cgdnd3Mtd2l6EAMYADICCAAyBggAEBYQHjoHCAAQRxCwAzoECAAQHjoJCAAQyQMQFhAeUNgZWKsiYOEoaAFwAngAgAGFAYgB1AaSAQM1LjOYAQCgAQGqAQdnd3Mtd2l6yAEHwAEB&sclient=gws-wiz) to see if the result is correct while avoid changing the timezone in the json file. You can change the open/close values in the corresponding day to test behavior locally. Also, you can add the current date with `MM/DD/YYYY` format to the holidays object to test it.

~WIP:~
- ~Add tests (in case this approach is approved).~
- ~Separate the `operatingHours` by channel. This can be done nesting the OperatingInfo -> operatingHours under channel key string (i.e. { "facebook": { "1": { open: xx, "close": yy }, ... },  "webchat": { ... }  }). This assumes that `holidays` and `timezone` will be shared across channels (what I believe, is correct). Please suggest if there's any other approach that makes more sense to you!~


assets/operatingInfo/za.private.json
```diff
@@ -0,0 +1,15 @@
...
+    "12/25/2021": "Christmas"
+  },
+  "operatingHours": {
+    "1": [{ "open": 9, "close": 11 }, { "open": 13, "close": 18 }],
+    "2": [{ "open": 9, "close": 11 }, { "open": 13, "close": 18 }],
+    "3": [{ "open": 9, "close": 11 }, { "open": 13, "close": 18 }],
+    "4": [{ "open": 9, "close": 11 }, { "open": 13, "close": 18 }],
+    "5": [{ "open": 9, "close": 11 }, { "open": 13, "close": 18 }],
+    "6": [],
+    "7": []
```
GPaoloni, line 13: Empty array means there are no open shifts for this day. This is helpful in the below expression of the function
`const isOpenNow = operatingHours[dayOfWeek].some(isInOpenShift);`
as an empty array will return `false` for any callback provided to `some` method.

functions/operatingHours.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+type EnvVars = {
+  OPERATING_INFO_KEY: string;
+};
+
+export type Body = {};
+
+const isOpen = (hour: number) => (shift: OperatingShift): boolean =>
+  hour >= shift.open && hour < shift.close;
```
nickhurlburt, line 37: This means we can't have opening hours that aren't on the hour, such as 8:30 - 11:30.  Would it be difficult to support that now?

nickhurlburt (approved):
Looks good to me, just with one question.

functions/operatingHours.ts
GPaoloni: No it is not, my last commit adds this!

functions/operatingHours.protected.ts
```diff
@@ -0,0 +1,101 @@
...
+
+type EnvVars = {
+  OPERATING_INFO_KEY: string;
+};
+
+export type Body = {
+  channel?: string;
+};
+
+const isOpen = (hour: number) => (shift: OperatingShift): boolean =>
```
nickhurlburt, line 38: Maybe name `hour` differently, since it isn't just the hour anymore.

functions/operatingHours.protected.ts
```diff
@@ -0,0 +1,101 @@
...
+      Runtime.getAssets()[`/operatingInfo/${OPERATING_INFO_KEY}.json`].open(),
+    );
+
+    if (!operatingInfo || !operatingInfo.operatingHours[channel])
+      throw new Error(
+        `Operating Info not found for channel ${channel}. Check OPERATING_INFO_KEY env vars and a matching OperatingInfo json file for it.`,
+      );
+
+    const { timezone, holidays, operatingHours } = operatingInfo;
+    const hour = parseInt(
```
nickhurlburt, line 71: Same as earlier comment, different name than `hour`

nickhurlburt (approved):
Just one naming thing.

PR #40, opened 2021-04-09
[Development only] Added operating info json and envar

GPaoloni:
This PR adds a json file with operating info for two channels: `webchat` is always open and `facebook` is always closed. The purpose of this is to allow the team to test the behavior of the function in our environment.

### DIFF
--- a/assets/operatingInfo/aselo-dev.json
+++ b/assets/operatingInfo/aselo-dev.json
@@ -5,13 +5,13 @@
   },
   "operatingHours": {
     "webchat": {
-      "1": [{ "open": 0, "close": 2359 }],
-      "2": [{ "open": 0, "close": 2359 }],
-      "3": [{ "open": 0, "close": 2359 }],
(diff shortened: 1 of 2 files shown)


assets/operatingInfo/aselo-dev.json
```diff
@@ -0,0 +1,26 @@
+{
+  "timezone": "America/New_York",
+  "holidays": { 
+    "12/25/2021": "Christmas"
+  },
+  "operatingHours": {
+    "webchat": {
+      "1": [{ "open": 0, "close": 2359 }],
```
nickhurlburt, line 8: If I remember right, the logic for close was a strict `<`, not `<=`.  So this may return `closed` at 11:59pm.

nickhurlburt (approved):
Looks good, with maybe one small issue

assets/operatingInfo/aselo-dev.json
GPaoloni: You are right, thanks for the catch!

PR #43, opened 2021-04-12
Fix .yaml file not including env var 

GPaoloni:
This PR fixes the GH action file for deploying to development. The variable was being loaded in the scope of the action but never pasted into the created `.env` file, thus it was not included in the serverless service.


GPaoloni:
Tagging all the team so you are aware of why env vars were failing for me after #42, but I'll go ahead and merge as the change is trivial.

PR #45, opened 2021-04-15
Added retry logic to assignOfflineContact

GPaoloni:
This PR added retry logic to `assignOfflineContact`.

### DIFF
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -32,6 +32,17 @@ type TaskInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Con
 // eslint-disable-next-line prettier/prettier
 type WorkerInstance = PromiseValue<ReturnType<ReturnType<ReturnType<ReturnType<Context['getTwilioClient']>['taskrouter']['workspaces']>['workers']>['fetch']>>;
 
+type AssignmentResult = {
+  type: 'error',
+  payload: { status: number, message: string, taskRemoved: boolean, attributes?: string }
+} |  { type: 'success', newTask: TaskInstance };
+
+const wait = (ms: number): Promise<void> => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+};
+
 const cleanUpTask = async (task: TaskInstance, message: string) => {
   const { attributes } = task;
   const taskRemoved = await task.remove();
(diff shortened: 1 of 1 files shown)


functions/assignOfflineContact.ts
```diff
@@ -45,12 +56,19 @@ const cleanUpTask = async (task: TaskInstance, message: string) => {
...
   newTask: TaskInstance,
-) => {
+  retry: number,
+): Promise<AssignmentResult> => {
   const reservations = await newTask.reservations().list();
   const reservation = reservations.find(r => r.workerSid === event.targetSid);
 
-  if (!reservation)
+  if (!reservation) {
+    if (retry < 3) {
```
nickhurlburt, line 65: Since we don't know how long this might actually take, maybe increase this to 5?  Hopefully it doesn't take more than a second.

functions/assignOfflineContact.ts
```diff
@@ -62,6 +80,9 @@ const assignToAvailableWorker = async (
   if (completed.reservationStatus !== 'completed')
     return cleanUpTask(newTask, 'Error: reservation for task not completed.');
 
+  // eslint-disable-next-line no-console
+  if (retry) console.warn(`Needed ${retry} retries to get reservation`);
```
nickhurlburt, line 84: Have you tested if this does produce output in the debugger or elsewhere?

functions/assignOfflineContact.ts
```diff
@@ -84,18 +105,13 @@ const assignToOfflineWorker = async (
     attributes: JSON.stringify({ ...previousAttributes, waitingOfflineContact: true }), // waitingOfflineContact is used to avoid other tasks to be assigned during this window of time (workflow rules)
   });
 
-  const result = await assignToAvailableWorker(event, newTask);
+  const result = await assignToAvailableWorker(event, newTask, 0);
```
nickhurlburt, line 108: Consider giving `retry` a default value of `0`, rather than passing it in here.  However, not a big deal to me.

nickhurlburt (approved):
Looks good -- some comments.  Let's discuss deployment at standup.  I'd like to get this deployed as soon as possible to ZM/ZA Prod, but we should also be careful.

functions/assignOfflineContact.ts
GPaoloni: Yes! It is displayed in the Twilio debugger error log.

Screenshot from 2021-04-15 12-55-12

PR #48, opened 2021-04-21
Deploy actions updated with operating info keys

GPaoloni:
This PR updates GH actions to include the corresponding key of each environment for operating hours file. Note that there is no ZM file, so calling the function without bundling the file first will result in an error.

Corresponding parameters already filled in AWS Parameter Store.

### DIFF
--- a/.github/workflows/test_staging.yml
+++ b/.github/workflows/test_staging.yml
@@ -59,6 +59,11 @@ jobs:
         with:
           ssm_parameter: "STG_TWILIO_TEST_WORKSPACE_SID"
           env_variable_name: "TWILIO_WORKSPACE_SID"
+      - name: Set operating information file key
+        uses: "marvinpinto/action-inject-ssm-secrets@latest"
+        with:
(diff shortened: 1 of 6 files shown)


murilovmachado (approved):
Looks good to me. The reason there's no ZM file is because they currently don't use operating hours?

GPaoloni:
@murilovmachado Right, the function is not called at all. If we ever add that behavior, we'll need to add `assets/operatingInfo/zm.private.json` file and it will work as expected.

.github/workflows/test_staging.yml
```diff
@@ -59,6 +59,11 @@ jobs:
         with:
           ssm_parameter: "STG_TWILIO_TEST_WORKSPACE_SID"
           env_variable_name: "TWILIO_WORKSPACE_SID"
+      - name: Set operating information file key
```
nickhurlburt, line 62: @alebarquin is this the temporary practice environment?  If so, better to label this something similar to development rather than staging, so we don't confuse it with other staging accounts.

nickhurlburt (approved):

PR #49, opened 2021-04-21
Twitter channel integration

GPaoloni:
Jira: 
- [Investigation](https://bugs.benetech.org/browse/CHI-511)
- [Flex -> Twitter hook](https://bugs.benetech.org/browse/CHI-615)
- [Twitter -> Flex hook](https://bugs.benetech.org/browse/CHI-616)
- [Flex - Twitter onUpdatedChannel webhook](https://bugs.benetech.org/browse/CHI-623)
- [Add tests](https://bugs.benetech.org/browse/CHI-668)
- [Secure public webhook](https://bugs.benetech.org/browse/CHI-674)
- [Include handle to show nice information on the UI](https://bugs.benetech.org/browse/CHI-678)

This PR bundles 3 webhooks that will be used to integrate Twitter DMs into Flex.


functions/webhooks/twitter/FlexToTwitter.protected.ts
```diff
@@ -0,0 +1,124 @@
...
+        console.log('Message sent from API call: ', Body);
+        resolve(success(TwitResponse));
+        return;
+      }
+
+      console.log('Message ignored (do not re-send self messages)');
+      resolve(success('Ignored event.'));
+    }
+
+    resolve(send(406)('Event Source not supported'));
```
andvirga, line 120: NIce, I haven't seen HTTP 406 (Not Accepted) for a while!

functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -0,0 +1,223 @@
...
+export type Body = {
+  crc_token?: string;
+  direct_message_events?: DirectMessageEvent[];
+  for_user_id?: string;
+  users?: { [key: string]: { name: string; screen_name: string } };
+};
+
+const twitterUniqueNamePrefix = 'twitter:';
+
+const validateWebhook = (crcToken: string, consumerSecret: string) => {
```
andvirga, line 48: Can you add a description of the parameters of this function?

andvirga (commented):
So far you're doing a great job! The code is clean and easy to follow!

functions/webhooks/twitter/FlexToTwitter.protected.ts
```diff
@@ -0,0 +1,127 @@
...
+    console.log('------ FlexToTwitter excecution ------');
+
+    if (event.Source === 'SDK') {
+      const TwitResponse = await sendTwitterMessage(context, recipientId, Body);
+      console.log('Message sent from SDK call: ', Body);
+      resolve(success(TwitResponse));
+      return;
+    }
+
+    if (event.Source === 'API' && event.EventType === 'onMessageSent') {
```
murilovmachado, line 92: I'm a little confused with `event.Source.` When it's value is `SDK` does it means it's a message that a counselor typed in herself, but when it's `API` it could mean it came from the autopilot? It's not clear to me which scenarios it's `SKD` or `API`.

functions/webhooks/twitter/FlexChannelUpdate.protected.ts
```diff
@@ -0,0 +1,80 @@
...
+        .fetch();
+
+      const { status } = JSON.parse(channel.attributes);
+
+      if (status === 'INACTIVE') {
+        await timeout(1000); // set small timeout just in case some cleanup is still going on
+        const removed = await channel.remove();
+        console.log(`INACTIVE channel with sid ${ChannelSid} removed: ${removed}`);
+        resolve(success(`INACTIVE channel removed: ${removed}`));
+        return;
```
murilovmachado, line 70: I think the name of the function could be more descriptive of what it actually does. Something like `RemoveInactiveFlexChannel.protected.ts`?

functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  await client.chat
+    .services(context.CHAT_SERVICE_SID)
+    .channels(channel.sid)
+    .update({
+      attributes: JSON.stringify({
+        ...channelAttributes,
+        channel_type: 'twitter',
+        twitterUserHandle: senderScreenName,
+        // customChannelType: 'twitter',
```
murilovmachado, line 125: is this commented because we won't need it?

functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -0,0 +1,270 @@
...
+  /* const onChannelUpdated = */
+  await client.chat
+    .services(context.CHAT_SERVICE_SID)
+    .channels(channel.sid)
+    .webhooks.create({
+      type: 'webhook',
+      configuration: {
+        method: 'POST',
+        url: `https://${context.DOMAIN_NAME}/webhooks/twitter/FlexChannelUpdate`,
+        filters: ['onChannelUpdated'],
```
murilovmachado, line 152: Who calls this function? Is it called only when we need to remove a Flex channel?

murilovmachado (approved):
Looks good to me, Gian! I've asked some questions in the code to help me understand better.

functions/webhooks/twitter/FlexToTwitter.protected.ts
GPaoloni: Yes, you get it right, counselor messages sent from Flex have SDK source, and every other message I've seen so far is API, including messages that the DM create in behalf of the client getting in contact (the child), so here we let them all go thru except the ones sent by the client, to avoid "redirecting self messages".

functions/webhooks/twitter/FlexChannelUpdate.protected.ts
GPaoloni: I'm about to change the behavior of this webhook, cause when doing this we lose the chat history in Twilio. The idea is to preserve the channels and keep track a map of Twitter user - active channel.

functions/webhooks/twitter/TwitterToFlex.ts
GPaoloni: Yes, it was a thing I was trying but is not needed in the end. Will remove it when doing cleanup, thanks for the catch!

functions/webhooks/twitter/TwitterToFlex.ts
GPaoloni: It is called on every channel update. In particular we are now interested when this channel is marked as `INACTIVE` to perform the appropriate cleanup.

PR #51, opened 2021-04-27
Added Zambia localization messages

andvirga:
Jira: [CHI-528](https://bugs.benetech.org/browse/CHI-528)

Primary reviewer: @GPaoloni 

Related to the following changes:
- [Flex-Plugins](https://github.com/tech-matters/flex-plugins/pull/428)
- [webchat](https://github.com/tech-matters/webchat/pull/28)


GPaoloni (approved):
Looks good! 
[Future work] - There is one issue that we may face in the future: right now the translations are not "scoped" in any way, so if two helplines want to have "Bemba" translations, we'll need to use a different name (as "Bemba" is already taken by ZM). We could scope translations with subfolders under `/assets/translations/` and pick the correct one with an envar. Minimal code changes will be required in `getMessages` and `getTranslations` functions.

PR #53, opened 2021-05-04
CHI-624 - HealthCheck for serverless

andvirga:
Jira: [CHI-624](https://bugs.benetech.org/browse/CHI-624)

Primary reviewer: @GPaoloni 

Changes description: 
- Created a new basic serverless function to perform a HealthCheck, if this function is unreachable or throws any code different than 200, curl will fail and therefore the whole CI process will fail.
- Added the serverless base URLs to AWS Parameter store.

### DIFF
--- a/.github/workflows/aselo_beta.yml
+++ b/.github/workflows/aselo_beta.yml
@@ -64,6 +64,11 @@ jobs:
         with:
           ssm_parameter: "PROD_TWILIO_AS_OPERATING_INFO_KEY"
           env_variable_name: "OPERATING_INFO_KEY"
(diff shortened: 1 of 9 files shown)


nickhurlburt (approved):
Looks good.

A second step to CHI-624 is to remove duplication.  This can be a separate PR.  Notice that you had to copy the same code into six different `yml` files.  We want to move to a situation where we only change this code in one place.  Perhaps the separate `yml`s just send an environment name (like `ASELO_DEV`) or an account id to a shared file that pulls all the right variables and runs the workflow.

PR #58, opened 2021-05-12
Moved offline contact task routing info to the function

GPaoloni:
This PR is related to https://github.com/techmatters/flex-plugins/pull/442.

This PR adds offline contact routing info into assignOfflineContact function.


nickhurlburt (approved):
This looks great to me.  If we've done some testing with it then I think it's good to go.

PR #54, opened 2021-05-06
CHI-624 - Created actions to reduce repeated code (DRY pattern)

andvirga:
JIra: [CHI-624](https://bugs.benetech.org/browse/CHI-624)

Changes description:
1. Created three actions to encapsulate common workflow logic:
- `compile-serverless:` builds `.env` file, executes `npm ci`, `npx tsc` and compiles the application.
- `deploy-twilio`: deploys the serverless app using twilio CLI.
- `check-serverless`: Executes the HealthCheck endpoint to verify that the deployed serverless is up and running.

2. Modified workflow to use these actions.
- Each action has a set of inputs and we're passing secrets values (Account SID, Token, etc.) to them.

More information:
I've followed the official [GitHub documentation](https://docs.github.com/en/enterprise-server@2.22/actions/creating-actions/about-actions) to create these actions. This is the maximum that I was able to reduce in the code, however, in the future, you may create one single workflow that triggers several actions to reduce even more the repeated steps. However, that would require a way to execute actions conditionally based on the value of the helpline (that could be provided by an extra input), so keep these in mind to continue improving this code.

This is my last contribution to Aselo and I wanted to mention that it was a great honor for me to had the chance of working on this amazing project! I really hope that Aselo keeps growing and getting implemented in new countries! WIsh you the best of luck!

PR #63, opened 2021-05-14
DRY staging actions

GPaoloni:
This PR
- Makes use of the centralized deployment action (added in #62) in the staging deployment workflows.
- Removes the unused actions (prototype in #54, added in #59).


.github/workflows/aselo_beta.yml
```diff
@@ -69,37 +69,17 @@ jobs:
         with:
           ssm_parameter: "ASELO_BETA_SERVERLESS_BASE_URL"
           env_variable_name: "SERVERLESS_BASE_URL"          
-      - name: Create environment variable file
```
nickhurlburt, line 72: A possible idea for the future:  we could move a lot of the code in lines 24-71 into the main action if we changed the variable names in AWS to use the account ids.  Then we just have to an account id variable for each.  For example, let's say we have this in `aselo_beta.yml`:
```
      ACCOUNT_SID: ${{ secrets.ASELO_BETA_ACCOUNT_SID }}
      # Call main-action to compile, deploy and check availability
      - name: Executing main-action
        uses: ./.github/actions/main-action
        with:
          account-sid: $ACCOUNT_ID
```
And we do a similar thing for each account yml file.  Then let's say that in AWS, rather than having a secret named `PROD_TWILIO_AS_API_KEY`, we have a secret named `SYNC_SERVICE_API_KEY_{{account sid}}`.  Then in the main action, we have:
```
     - name: Set Twilio Sync API key
        with:
          ssm_parameter: format("SYNC_SERVICE_API_KEY{0}", ${{ account-sid }})
          env_variable_name: "SYNC_SERVICE_API_KEY"
     ....
```
and so on for the other secrets.  Then all of that can go into the main action and we have less code in the workflow.

No need to do this now, and we'd have to be careful about changing names for items in the parameter store.  But an idea for later.

nickhurlburt (approved):
Looks great.  I left one comment for a future idea.

When we make changes like this, we should deploy related environments pretty soon after merging.  As long as there are no other big changes on serverless, I don't think those deploys need to be part of a big deploy meeting.

murilovmachado (approved):

PR #68, opened 2021-05-24
Added console.error on every catch block

GPaoloni:
This PR adds a `console.error` statement on catch block for every function so we can gather more context and useful information whenever an error occurs, in contrast to the current behavior that's not very helpful.
Screenshot from 2021-05-24 16-45-34
Screenshot from 2021-05-24 16-56-34

The new statement will output something like
Screenshot from 2021-05-24 16-56-05
Screenshot from 2021-05-24 16-56-20


nickhurlburt (approved):
Looks good!

murilovmachado (approved):
Nice!

PR #70, opened 2021-06-01
Fixed operatingHours function (Sunday bug)

GPaoloni:
This PR fixes a bug that occurs only on Sundays, spotted in ZA production. The error was caused because depending on the timezone, the "day of the week" was 0 or 7 for Sundays (and we only include "7" as key of day in the operating info json files).
Screenshot from 2021-06-01 15-45-33

This PR fixes the bug by using [moment.isoWeekDay](https://momentjs.com/docs/#/get-set/iso-weekday/) function instead, and includes test to confirm is working as expected.


nickhurlburt (approved):
Looks good!  Thanks for updating tests!

dee-luo (approved):

PR #71, opened 2021-06-04
Change from attribute when creating a new voice task

murilovmachado:
Every time a task is created, before going directly to Flex, it nows goes through `voiceWebhook` serverless function. If it's not a voice task, nothing is changed. If it's voice, this function will update the task `from` attribute value, making it unique. This is done by appending the `taskSid` onto the `from` attribute value.

**Deployment Plan**
1. Check that Insights behaves as expected;
2. Deploy to all other environments;
3. Setup VoiceWebhook for each account

**How Setup VoiceWebhook for each account?**
1. Deploy `serverless`
2. Copy function URL
image
3. Open TaskRouter Settings
image
4. Set Event Callback URL to be the function URL, and mark only **Task Created** Event under **Specific Events**
image

### DIFF
--- a/functions/voiceWebhook.protected.ts
+++ b/functions/voiceWebhook.protected.ts
@@ -1,0 +1,67 @@
+import '@twilio-labs/serverless-runtime-types';
+import { Context, ServerlessCallback } from '@twilio-labs/serverless-runtime-types/types';
+import {
+  responseWithCors,
+  bindResolve,
+  success,
+  error400,
+  error500,
+} from '@tech-matters/serverless-helpers';
+
+export type Body = {
+  EventType: string;
+  TaskSid: string;
+  TaskChannelUniqueName: string;
+};
+
+type EnvVars = {
+  TWILIO_WORKSPACE_SID: string;
+};
+
+const TASK_CREATED_EVENT = 'task.created';
+const VOICE_TASK_CHANNEL = 'voice';
+
+export const handler = async (
+  context: Context<EnvVars>,
+  event: Body,
+  callback: ServerlessCallback,
+) => {
+  const response = responseWithCors();
+  const resolve = bindResolve(callback)(response);
+
+  const { EventType, TaskSid, TaskChannelUniqueName } = event;
+
+  const isNewVoiceTask =
+    EventType === TASK_CREATED_EVENT && TaskChannelUniqueName === VOICE_TASK_CHANNEL;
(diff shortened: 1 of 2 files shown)


GPaoloni (approved):
Looks very good! :rocket:

nickhurlburt (commented):
A couple questions:
1. Does this affect any of our existing Insights reports, especially those that we currently display to customers?  That is, does changing `from` mean that the phone number in various reports will now show phone number + task id?  Do the fields that automatically read the phone number (like Country, State, etc) still work?  Could be helpful to get @dee-luo or @Humairaa127's help to look at that too.
2. This will require an additional deploy step of adding the voice webhook to accounts. 
 Can you record something in the PR description about how and when this should be done? 
 Should we do it right away so we don't forget, or should we wait for this to be deployed everywhere?
3. Please add the step of adding this callback to the Taskrouter config to the [Twilio account setup doc](https://benetech.app.box.com/file/772893818136) so we don't forget for future new accounts.

murilovmachado:
> 1. Does this affect any of our existing Insights reports, especially those that we currently display to customers?  That is, does changing `from` mean that the phone number in various reports will now show phone number + task id?  Do the fields that automatically read the phone number (like Country, State, etc) still work?  Could be helpful to get @dee-luo or @Humairaa127's help to look at that too.

This should **not** affect Insights reports. **From** should be a **unique** value for each task (as it currently is for all other channels). At the front-end code, we're fetching **defaultFrom** to get the phone number, which is different than the **from** attribute this PR changes. If somehow we're seeing +123456_WTXXX as phone number in insights reports, we're probably looking at the wrong attributes.

> 2. This will require an additional deploy step of adding the voice webhook to accounts.
>    Can you record something in the PR description about how and when this should be done?
>    Should we do it right away so we don't forget, or should we wait for this to be deployed everywhere?

Let's have this on development for some days, and have Dee, Humairaa or Joan double-check that Insights works as expected. If they're fine with it, I suggest deploying and configuring it for all environments.

> 3. Please add the step of adding this callback to the Taskrouter config to the [Twilio account setup doc](https://benetech.app.box.com/file/772893818136) so we don't forget for future new accounts.

I've written in this PR the steps to configure this. I'll add it to the docs as well.

murilovmachado:
@nickhurlburt @GPaoloni I've reverted master to not include this PR.

PR #74, opened 2021-06-17
Testing approach to support multiple Twitter accounts

GPaoloni:
This PR includes an extra step to the deployment script in an attempt to support multiple Twitter accounts for a single Twilio account.


GPaoloni:
@nickhurlburt I'm merging this to test if it works, but asking for review so you are aware of the approach I'm trying to support this.

PR #79, opened 2021-06-22
Preserve chat history for Twitter

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-676

Prior to this changes, when a Twitter channel contact was ended, the Programmable Chat Channel resource was deleted. 
This PR Preserve chat history (not deleting the Programmable Chat Channel resource). To accomplish this we use the following strategy:
- When a new Twitter DM arrives, we check if there is already an active channel for this Twitter user ID.
- If there's not active channel, we create a new channel, and save a new Sync document, named `'twitter:<userID>'` that contains the recently created channel sid (as `activeChannelSid` property).
- If there's an active channel, all subsequent messages will find it's sid mentioned above after a lookup to this Sync document.
- When the contact is over, the Sync document is removed.

The decision of making a new document per Twitter user is due to the [Sync limits](https://www.twilio.com/docs/sync/limits#sync-activity-limits), given that we can perform up to 20 reads per second per document, but have no limit in the amount of objects we can create (as long as we create at most 20 per second, which is a safe number for our use case I believe).

WIP:
- Fix CI
- Cleanup (leaving `console.log`s for now so you can see them in the Twilio console, in the dev environment).

### DIFF
--- a/functions/webhooks/twitter/TwitterToFlex.ts
+++ b/functions/webhooks/twitter/TwitterToFlex.ts
@@ -12,6 +12,7 @@ import crypto from 'crypto';
 type EnvVars = {
   ACCOUNT_SID: string;
   AUTH_TOKEN: string;
+  SYNC_SERVICE_SID: string;
   TWITTER_CONSUMER_KEY: string;
   TWITTER_CONSUMER_SECRET: string;
   TWITTER_ACCESS_TOKEN: string;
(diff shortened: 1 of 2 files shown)


functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -74,12 +75,43 @@ const isValidTwitterPayload = (event: Body, consumerSecret: string): boolean =>
 /**
  * Retrieves a channel by sid or uniqueName
  */
-const retrieveChannel = (context: Context<EnvVars>, uniqueName: string) =>
-  context
+const retrieveChannel = async (
+  context: Context<EnvVars>,
+  uniqueSenderName: string,
+): Promise<string | undefined> => {
```
murilovmachado, line 81: I believe it's now retrieving a channelSid, instead of a channel, right? if that's the case, I suggest changing the function name to be `retrieveChannelSid` and update the function comment above.

functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -74,12 +75,43 @@ const isValidTwitterPayload = (event: Body, consumerSecret: string): boolean =>
...
+  const userChannelMap = await context
     .getTwilioClient()
-    .chat.services(context.CHAT_SERVICE_SID)
-    .channels(uniqueName)
-    .fetch();
+    .sync.services(context.SYNC_SERVICE_SID)
+    .documents.create({
+      data: { activeChannelSid: channelSid },
+      uniqueName: uniqueSenderName,
+      ttl: 259200, // auto romeoved after 3 days
```
murilovmachado, line 110: Typo in `remeoved`. Another question: why 3 days? Is it related to how long we want to keep the task/channel open?

functions/webhooks/twitter/TwitterToFlex.ts
```diff
@@ -179,33 +209,33 @@ const sendChatMessage = async (
...
-      channelSid = newChannel.sid;
-    } catch (err2) {
-      throw new Error(`Error while creating the new channel ${err2.message}`);
+
+      channelSid = newChannelSid;
+      await createUserChannelMap(context, uniqueSenderName, channelSid);
     }
+  } catch (err) {
+    throw new Error(`Error while creating the new channel ${err.message}`);
   }
```
murilovmachado, line 239: If for some reason `createUserChannelMap` fails, we will have a "hanging" channel (that was just created above). Is that a problem? Should we try to remove this "hanging" channel on the catch/finally block?

murilovmachado (approved):
Looks good to me! I didn't know these sync services existed. I had a couple of questions/suggestions.

functions/webhooks/twitter/TwitterToFlex.ts
GPaoloni: Thanks! The 3 days timeout is an arbitrary based on the max time we've been asked a task should live so far. This was one of the questions I had. I believe we need some timeout (in case we miss the "clean up" event), but don't know what would be the correct number. Maybe @nickhurlburt have an opinion?

nickhurlburt: I believe that currently, a task will be timed out after 24 hours.  But it can be set to survive for up to 2 weeks.  If there's no harm to doing so, I'd say make it 2 weeks.  When we add email it'll have a longer TTL, and we won't have to remember to update this.

nickhurlburt (approved):
Looks good to me, with Murilo's comments.  One question though:  do we know how we can access the conversation history later?  For most text-based channels we can use the messaging API, but I'm guessing that won't work here?

GPaoloni:
@nickhurlburt As already discussed via Slack, we can retrieve the channel using the same method as Webchat, and for that we need to start saving the channel sid related to a contact.

functions/webhooks/twitter/TwitterToFlex.ts
GPaoloni: I don't think there is any problem but already added a cleanup for it.

PR #80, opened 2021-07-08
Added getWorkerAttributes function

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-719
This PR is related to https://github.com/techmatters/flex-plugins/pull/472

Adds a new function that given a worker sid, retrieves it's attributes.

### DIFF
--- a/functions/getWorkerAttributes.ts
+++ b/functions/getWorkerAttributes.ts
@@ -1,0 +1,58 @@
+import '@twilio-labs/serverless-runtime-types';
+import {
+  Context,
+  ServerlessCallback,
+  ServerlessFunctionSignature,
+} from '@twilio-labs/serverless-runtime-types/types';
+import {
+  responseWithCors,
+  bindResolve,
+  error400,
+  error500,
+  success,
+} from '@tech-matters/serverless-helpers';
+
+const TokenValidator = require('twilio-flex-token-validator').functionValidator;
+
+type EnvVars = {
+  TWILIO_WORKSPACE_SID: string;
+};
+
+export type Body = {
+  workerSid?: string;
+};
+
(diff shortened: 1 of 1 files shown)


functions/getWorkerAttributes.ts
```diff
@@ -0,0 +1,58 @@
...
+        return;
+      }
+
+      const client = context.getTwilioClient();
+      const worker = await client.taskrouter
+        .workspaces(context.TWILIO_WORKSPACE_SID)
+        .workers(workerSid)
+        .fetch();
+
+      const workerAttributes = JSON.parse(worker.attributes);
```
nickhurlburt, line 44: I think this is too much of a security issue.  This gives any counselor, regardless of their permissions, access to all the attributes of another counselor.  In a large organization with hundreds of volunteers, that's way too much data.  Especially since we might add more attributes in the future and forget that they're coming through here.

Let's just whitelist the `helpline` attribute and return that only (but we should return it inside an object so that we can possibly add more attributes later.

nickhurlburt (commented):
Consider this approved, but make sure to change it to restrict to just returning essential attributes only (probably just the `helpline`).

Please go ahead and deploy to dev and staging tomorrow.

functions/getWorkerAttributes.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      if (workerAttributes.helpline === undefined)
+        throw new Error(
+          'Error: the target worker does not have helpline attribute set, check the worker configuration.',
+        );
```
murilovmachado, line 49: Do we really want a 500 error in this case?

murilovmachado (commented):
Looks good to me! Just one question on a 500 error.

PR #82, opened 2021-07-14
FlexChannelUpdate false-positive error fix

GPaoloni:
This PR fixes an error @Humairaa127 spotted in Zambia. Once a Twitter task is completed, the FlexChannelUpdate webhook might be called several times, but we wan't to cleanup the user-channel map (Sync Document) only once (to avoid generating error logs that are caused because of trying to "delete something that was already deleted").

### DIFF
--- a/functions/webhooks/twitter/FlexChannelUpdate.protected.ts
+++ b/functions/webhooks/twitter/FlexChannelUpdate.protected.ts
@@ -82,7 +82,7 @@ export const handler = async (
       const { status, from } = JSON.parse(channel.attributes);
 
       if (status === 'INACTIVE') {
-        await timeout(3000); // set small timeout just in case some cleanup is still going on
+        await timeout(1000); // set small timeout just in case some cleanup is still going on
 
         const removed = await cleanupUserChannelMap(context, from);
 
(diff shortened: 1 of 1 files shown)


functions/webhooks/twitter/FlexChannelUpdate.protected.ts
```diff
@@ -66,11 +84,7 @@ export const handler = async (
...
         await timeout(1000); // set small timeout just in case some cleanup is still going on
 
-        const removed = await context
-          .getTwilioClient()
-          .sync.services(context.SYNC_SERVICE_SID)
-          .documents(from)
-          .remove();
+        const removed = await cleanupUserChannelMap(context, from);
 
         console.log(`INACTIVE channel triggered map removal for ${from}, removed ${removed}`);
```
nickhurlburt, line 90: For the line below, I'm not sure what the value of `removed` is if it did remove a document, but the message if it failed to remove will be:
`INACTIVE channel triggered map removal for <whatever 'from' is>, removed false`
Is "removed false" the best message?  Is there a message that's more clear?

nickhurlburt (approved):
Looks good!

PR #85, opened 2021-07-28
Added addCustomerExternalId

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-742

This task adds external id into task attributes-> customer object when a new task is created. I've confirmed that the attribute is properly set as soon as the task start existing in Taskrouter, however I have not tested if this fixes the reported Insights bug. The function is deployed in Aselo dev and already linkend so we can test the behavior even before merging the PR (I just don't know how @dee-luo).

To make this work, after the function is deployed in the proper environment, we also need to specify the url as the "Taskrouter callback url", in Twilio console under TaskRouter -> Flex Task Assignment -> Settings -> Event Callbacks.

WIP:
- Unit tests.


murilovmachado (approved):
Looks good to me!

PR #87, opened 2021-08-11
Post contact survey: proof of concept for chat contacts

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-771
This PR is related to https://github.com/techmatters/flex-plugins/pull/490.

This PR adds two functions to work trigger a post contact survey:
- `postSurveyInit` which adds a new webhook to the chat channel pointing to a chat bot and triggers it by sending a message.
- `postSurveyComplete` that removes the above bot and mimic the behavior of the channel janitor.

### DIFF
--- a/.github/actions/beta-action/action.yml
+++ b/.github/actions/beta-action/action.yml
@@ -1,0 +1,81 @@
+# Install the Twilio CLI and the serverless plugin then deploy the function
+name: 'main-action'
+description: 'Install the Twilio CLI and the serverless plugin then deploy the function'
+inputs:
+  account-sid:
+    description: 'Twilio Account SID'
+    required: true
+  auth-token:
+    description: 'Twilio Token'
+    required: true
+  workspace-sid:  
+      description: 'Twilio Workspace SID'
+      required: true
+  transfer-workflow-sid:  
+    description: 'Chat Transfer Workflow SID'
+    required: true
+  sync-service-api-key:  
+    description: 'Sync Service API Key'
+    required: true
+  sync-service-api-secret:  
+    description: 'Sync Service API Secret'
+    required: true
+  sync-service-sid:
+    description: 'Sync Service SID'
+    required: true
+  chat-service-sid:
+    description: 'Chat Service SID'
+    required: true
+  flex-proxy-service-sid:
+    description: 'Flex Proxy Service SID'
+    required: true
+  operating-info-key:
+    description: 'Operating Info Key'
+    required: true
+  serverless-base-url:
+    description: 'Serverless base url'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    # Prepare everything to compile the serverless application with proper env
+    - name: Create environment variable file
+      run: touch .env
+      shell: bash
+    - name: Fill .env
+      run: |
+          cat <<EOT >> .env
+          TWILIO_WORKSPACE_SID=${{ inputs.workspace-sid }}
+          TWILIO_CHAT_TRANSFER_WORKFLOW_SID=${{ inputs.transfer-workflow-sid }}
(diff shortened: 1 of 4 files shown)


functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,109 @@
...
+  context: Context<EnvVars>,
+  event: Event,
+  callback: ServerlessCallback,
+) => {
+  Object.entries(event).forEach(([k, v]) => {
+    console.log(k, JSON.stringify(v));
+  });
+  try {
+    const memory = JSON.parse(event.Memory);
+    const { ServiceSid, ChannelSid } = memory.twilio.chat;
```
murilovmachado, line 77: Does `event.memory` holds the answers from the post-survey?
If yes, should we send them somewhere else? (HRM/Insights)

murilovmachado (approved):

functions/postSurveyComplete.protected.ts
GPaoloni: Yes to both questions, but for now I just wanted to show how we can do a post contact chat with the bot and then use the memory. Now we need to create a task (so it goes to Inishts) and sotre this somewhere in Hrm too.

nickhurlburt (approved):
I looked through very quickly... looks like a good approach.  As mentioned in the other PR, we need to make sure we don't miss cleanup.  Have we checked if it's possible to trigger the janitor ourselves?

GPaoloni:
@nickhurlburt I don't think we can manually trigger the Janitor. I'm also worried that something goes wrong here, but right now can't think of other approach for the post surveys. 
@nickhurlburt @murilovmachado I've included one more commit where I change the deploy script (and add a new `beta-action`, so all the rest of the scripts can continue to work the same). Let me know if you guys think we can/makes sense merge this.

nickhurlburt:
@GPaoloni  What is the behavior if this gets deployed to other accounts?  Will it have any effect if it is deployed on its own, or does it need feature flags on the plugin to be invoked?

If it's safe for this to be deployed to other accounts, I have no problem with merging it.

.github/actions/beta-action/action.yml
```diff
@@ -0,0 +1,81 @@
...
+    required: true
+  sync-service-sid:
+    description: 'Sync Service SID'
+    required: true
+  chat-service-sid:
+    description: 'Chat Service SID'
+    required: true
+  flex-proxy-service-sid:
+    description: 'Flex Proxy Service SID'
+    required: true
```
murilovmachado, line 31: This is the only difference in this new script, right?

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,111 @@
...
+
+type EnvVars = {
+  FLEX_PROXY_SERVICE_SID: string;
+};
+
+const deleteProxySession = async (context: Context<EnvVars>, proxySession: string) => {
+  try {
+    const client = context.getTwilioClient();
+    const ps = await client.proxy
+      .services(context.FLEX_PROXY_SERVICE_SID)
```
murilovmachado, line 22: Even though `staging` and `production` are not using `beta-action`, this function would be deployed, correct?
Is there any chance this function would be called on `staging` or `production`? (because if so, `context.FLEX_PROXY_SERVICE_SID` would be undefined)

murilovmachado (approved):
Looks good to me. I've left a few questions, just to make sure it's safe to deploy.

functions/postSurveyComplete.protected.ts
GPaoloni: Yes it will be deployed but no, it should never be invoked. The breaking change is in the task router callback tho.

GPaoloni:

PR #90, opened 2021-08-20
[BREAKING CHANGE] Mimic channel janitor behavior on survey task complete

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-756.
This PR is related tohttps://github.com/techmatters/flex-plugins/pull/494

This PR uses taskrouter to bind a channel used for survey to a task (so it will be cleaned up no matter the results of the bot).
- `postSurvetInit` is hit by the counselor (from Flex) when the original task is finished. Here we create a new task and trigger the post survey.
- `postSurveyComplete` is called when the bot execution finishes, and here is where we should edit the task attributes to save in Insights and also save in HRM.
- `postSurveyJanitor` mimics the behavior of the channel janitor: deletes the proxy session if any and then deactivates the chat channel.

NOTES: 
- The changes on `package.lock` were triggered by npm. No new dependency added,
- The functions under `/private` folder are for internal use (taskrouter callback), but they are `.protected.` files as for some reason I can't deploy `.private.` ones. Just mentioning this a trying to hit one of them via HTTP call will fail.
- Deploying this requires updating the taskrouter callback url (to be `/webhooks/taskrouterCallback`).
- This approach is taken from [this Twilio guide](https://www.twilio.com/blog/post-task-surveys-with-flex-insights).
- For this approach to work as expected we must do the following work in taskrouter (examples can be found in Aselo dev): 
  - Create Survey Task Queue.
  - Create Survey Workflow.
  - Create Survey Task Channel (not sure if this is really needed).
  - Add task canceled and completed events to be sent to the taskrouter callback and point to the correct url.
  - Disable the channel janitor for the chat channels.

WIP:
- Move out hardcoded SIDs.
- Fix CI.
- Write unit tests?


nickhurlburt (approved):
I didn't look in great detail, but overall makes sense!  Unit tests would be a good idea :)

functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const baseUrl = `https://${context.DOMAIN_NAME}`;
+
+  try {
+    const { EventType } = event;
+
+    if (EventType === TASK_CREATED_EVENT) {
+      const handlerPath = Runtime.getFunctions().addCustomerExternalId.path;
+      const addCustomerExternalId = require(handlerPath).addCustomerExternalId as AddCustomerExternalId;
+      await addCustomerExternalId(context, event);
```
murilovmachado, line 50: I remember we're were having issues like `calling 'path' of undefined`.

If this is still a potential issue, I feel like this `taskrouterCallback` could be split into two separate callbacks to avoid this workaround to invoke other functions `Runtime.getFunctions().addCustomerExternalId.path`:

1. TASK_CREATED_EVENT:  addCustomerExternalId
2. TASK_CANCELED_EVENT || TASK_COMPLETED_EVENT: postSurveyJanitor

If that's not an issue anymore, nevermind.

murilovmachado (approved):
Looks good. Just one comment on the taskrouter callback.

GPaoloni:

PR #91, opened 2021-08-26
Grab values from bot memory and drop them in corresponding task attributes

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-781

This PR adds insights services like the ones we use in Flex to fill in the task attributes (and save the data in Insights).
The idea is that we provide a JSON file with the same format as the "one to many specs" file for Flex, but instead of paths we have a list of "question names", where each question name is the same as it appears defined in the post survey autopilot assistant.

As future work, we should integrate this with the actual form definition version used in Flex.

Notes:
- The post survey specs for `v1` form definition is just an example. We should be aware before deploying this as it can save garbage data into insights.


assets/formDefinitions/v1/insights/postSurvey.private.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "insightsObject": "customers",
+    "attributeName": "customer_attribute_9",
+    "questions": ["question"]
+  },
+  {
+    "insightsObject": "customers",
+    "attributeName": "customer_attribute_10",
+    "questions": ["question"]
```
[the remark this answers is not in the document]
nickhurlburt: I'll leave it to you.  I agree that 'questions' is a better name than 'path' in this situation.  If we ever merge this into the the other insights config files, we might have to change the names to be the same.

functions/helpers/insightsService.private.ts
```diff
@@ -0,0 +1,89 @@
...
+  };
+};
+
+const applyCustomUpdate = (customUpdate: OneToManyConfigSpec): SurveyInsightsUpdateFunction => {
+  return memory => {
+    const updatePaths = customUpdate.questions.map(
+      question => `twilio.collected_data.collect_survey.${question}.answer`, // Path where the answer for each question should be in bot memory
+    );
+    // concatenate the values, taken from dataSource using paths (e.g. 'contactForm.childInformation.province')
+    const value = updatePaths.map(path => get(memory, path, '')).join(delimiter);
```
murilovmachado, line 52: This result `value` is all the answers as a single string separated by `;`?

functions/helpers/insightsService.private.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return customUpdatesFuns;
+};
+
+export const buildSurveyInsightsData = (oneToManyConfigSpecs: OneToManyConfigSpecs) => (
+  taskAttributes: TaskAttributes,
+  memory: BotMemory,
+) => {
+  // NOTE: I assume that if surveys are enabled this is not needed, right?
+  // if (!shouldSendInsightsData(task)) return previousAttributes;
```
murilovmachado, line 75: It makes sense that if there's no post survey, it would just return the `previousAttributes`.
(I'm just not sure how `shouldSendInsightsData(task)` works)

murilovmachado (approved):
Looks good!

GPaoloni:

PR #93, opened 2021-09-01
Added support for storing survey data in hrm

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-790

This PR makes use of the API exposed in https://github.com/techmatters/hrm/pull/122 to save post survey results in HRM.

Notes:
- Using `axios` as the library to make HTTP requests as some issues appears when using `node-fetch`.
- A system key needs to be created for this.

~WIP:~
- ~Fix CI~


functions/postSurveyComplete.protected.ts
```diff
@@ -44,29 +97,28 @@ export const handler: ServerlessFunctionSignature<EnvVars, Event> = async (
...
-          const taskAttributes = JSON.parse(surveyTask.attributes);
-          const finalAttributes = buildSurveyInsightsData(postSurveyConfigSpecs)(taskAttributes, memory);
-          console.log('finalAttributes: ', JSON.stringify(finalAttributes));
+          const postSurveyConfigSpecs = JSON.parse(postSurveyConfigJson.open()) as OneToManyConfigSpecs;
 
-          await surveyTask.update({ attributes: JSON.stringify(finalAttributes) }); 
+          // parallel execution to save survey collected data in insights and hrm
+          await Promise.all([
+            saveSurveyInInsights(postSurveyConfigSpecs, memory, surveyTask),
+            saveSurveyInHRM(postSurveyConfigSpecs, memory, surveyTask, hrmBaseUrl, context.SYSTEM_SECRET),
```
nickhurlburt, line 118: Note that it's possible that one of these succeeds and the other fails.  That then means that we have an inconsistency between the two.  When storing contacts, I don't think this is okay.  But for post-surveys, which are less important, this is fine.  But we should still eventually add post-surveys to our Insights/HRM audit checker.

functions/postSurveyComplete.protected.ts
```diff
@@ -17,6 +33,45 @@ export interface Event {
...
+
+  const data = buildDataObject(postSurveyConfigJson, memory);
+
+  const body: PostSurveyBody = {
+    contactTaskId: taskAttributes.contactTaskId,
+    taskId: surveyTask.sid,
+    data
+  };
+
+  await axios({
```
nickhurlburt, line 66: Can you add a comment here as to why axios is required instead of fetch?

nickhurlburt (approved):
I mainly reviewed for security.  Good for @murilovmachado also to take a look.

functions/postSurveyComplete.protected.ts
murilovmachado: I agree with Nick here

murilovmachado (approved):
Looks good!

GPaoloni:

PR #94, opened 2021-09-06
Move the remaining service SIDs to the env vars (deployment scripts)

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-788

This PR is about cleaning the hardcoded resources that will change from one account to another. They are added to the "beta deployment script" in order to preserve the deployments to other environments without breaking them. 

As an important note, even if this is preserving the deployment script untouched, there are still breaking changes regarding the Taskrouter Event Callback introduced in https://github.com/techmatters/serverless/pull/90.

### DIFF
--- a/.github/actions/beta-action/action.yml
+++ b/.github/actions/beta-action/action.yml
@@ -35,8 +35,8 @@ inputs:
   flex-proxy-service-sid:
     description: 'Flex Proxy Service SID'
     required: true
-  system-secret:
-    description: 'Secret to make HRM system calls'
+  hrm-static-key:
+    description: 'HRM static key to make HRM calls'
     required: true
   survey-workflow-sid:
     description: 'Survey Workflow SID'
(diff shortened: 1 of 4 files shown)


functions/postSurveyInit.ts
```diff
@@ -36,7 +38,7 @@ const createSurveyTask = async (context: Context<EnvVars>, event: Required<Body>
   };
 
   const surveyTask = await client.taskrouter.workspaces(context.TWILIO_WORKSPACE_SID).tasks.create({
-    workflowSid: 'WW6a967d8f663083bdb8ae586539fa71d7', // TODO: move this out
```
nickhurlburt, line 39: While this and the values on Line 73 aren't really secrets, it would still be preferable to keep them out of a public repo completely.

.github/actions/beta-action/action.yml
```diff
@@ -26,15 +26,24 @@ inputs:
...
   operating-info-key:
     description: 'Operating Info Key'
     required: true
   serverless-base-url:
     description: 'Serverless base url'
     required: true
+  flex-proxy-service-sid:
+    description: 'Flex Proxy Service SID'
+    required: true
+  system-secret:
```
nickhurlburt, line 38: As with comments on https://github.com/techmatters/hrm/pull/122, this is too general.  Maybe `api-key-write-only`?  I don't like that either, but I want to be clear about what this is.

nickhurlburt (approved):

.github/actions/beta-action/action.yml
murilovmachado, line 38: The name `system-secret` is not very clear. Better change it for `hrm-secret`?

murilovmachado (approved):
Looks Good!

GPaoloni:

PR #95, opened 2021-09-06
[BREAKING CHANGE] Post Survey feature branch

GPaoloni:
This PR is about adding post survey support from the Serverless side. It is the entire bundle of the following (open) PRs:
- https://github.com/techmatters/serverless/pull/87
- https://github.com/techmatters/serverless/pull/90
- https://github.com/techmatters/serverless/pull/91
- https://github.com/techmatters/serverless/pull/93
- https://github.com/techmatters/serverless/pull/94

but showing the diff against master so it's easier to give a full review of all the code included to support this feature.

WIP:
- Documentation, notes before merging & deploying, setup process.
- Add unit tests?


GPaoloni:
@nickhurlburt @murilovmachado This is the entire feature PR for post surveys. I've closed all the intermediate PRs as it makes more sense to just have the final code reviewed, but you can see the comments in previous PRs and check that they have been addressed. Please review this one when you have some time, but no rush doing so, as it's a big one. I'm now in the process of writing some documentation to help understand the workflow and how this interacts altogether.

functions/helpers/postSurveyJanitor.private.ts
```diff
@@ -0,0 +1,80 @@
+// eslint-disable-next-line prettier/prettier
```
nickhurlburt, line 1: It's worth adding some comments to this file also, to be clear about what it's replacing and why that's necessary.

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+  });
+};
+
+export const handler: ServerlessFunctionSignature<EnvVars, Event> = async (
+  context: Context<EnvVars>,
+  event: Event,
+  callback: ServerlessCallback,
+) => {
+  console.log('-------- postSurveyComplete execution --------');
+  Object.entries(event).forEach(([k, v]) => console.log(k, JSON.stringify(v)));
```
nickhurlburt, line 83: I assume that some of these console statements will disappear before we deploy, at least to production?

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (definitionVersion && postSurveyConfigJson && postSurveyConfigJson.open) {
+          const postSurveyConfigSpecs = JSON.parse(postSurveyConfigJson.open()) as OneToManyConfigSpecs;
+
+          // parallel execution to save survey collected data in insights and hrm
+          await Promise.all([
+            saveSurveyInInsights(postSurveyConfigSpecs, memory, surveyTask),
+            saveSurveyInHRM(postSurveyConfigSpecs, memory, surveyTask, hrmBaseUrl, context.HRM_STATIC_KEY),
+          ]);
+        } else {
+        // eslint-disable-next-line no-console
```
nickhurlburt, line 121: nit: indent

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const data = buildDataObject(postSurveyConfigJson, memory);
+
+  const body: PostSurveyBody = {
+    contactTaskId: taskAttributes.contactTaskId,
+    taskId: surveyTask.sid,
+    data
+  };
+
+  await axios({
```
nickhurlburt, line 66: Please add a comment explaining why we're using axios, especially since most of our code does not.

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+  HRM_STATIC_KEY: string;
+};
+
+const saveSurveyInInsights = async (postSurveyConfigJson: OneToManyConfigSpecs, memory: BotMemory, surveyTask: TaskInstance) => {
+  const handlerPath = Runtime.getFunctions()['helpers/insightsService'].path;
+  const buildSurveyInsightsData = require(handlerPath)
+    .buildSurveyInsightsData as BuildSurveyInsightsData;
+
+  const taskAttributes = JSON.parse(surveyTask.attributes);
+  const finalAttributes = buildSurveyInsightsData(postSurveyConfigJson)(taskAttributes, memory);
```
nickhurlburt, line 45: Why is this function curried?  Especially when considering that the function in L58 is not curried.

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+          const postSurveyConfigSpecs = JSON.parse(postSurveyConfigJson.open()) as OneToManyConfigSpecs;
+
+          // parallel execution to save survey collected data in insights and hrm
+          await Promise.all([
+            saveSurveyInInsights(postSurveyConfigSpecs, memory, surveyTask),
+            saveSurveyInHRM(postSurveyConfigSpecs, memory, surveyTask, hrmBaseUrl, context.HRM_STATIC_KEY),
+          ]);
+        } else {
+        // eslint-disable-next-line no-console
+          console.error('Missing definition for post survey. Not saving to insights.');
```
nickhurlburt, line 122: This is a confusing error because you say 'Not saving to insights' but you are not saving to insights OR to hrm.

Can you add some debug information to this, so that if we get paged due to this error, we can debug it?

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        await surveyTask.update({ assignmentStatus: 'canceled' }); // can't complete a pending task only cancel it
```
nickhurlburt, line 125: This comment is not clear.  Will this function only be called for "pending" tasks?

functions/postSurveyComplete.protected.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+
+    const actions = [
```
nickhurlburt, line 129: Maybe comment that this is the message being sent back to the person, if that's what it is.  Do we need to use some kind of variable so that we can localize the message?

functions/postSurveyInit.ts
```diff
@@ -0,0 +1,112 @@
...
+        method: 'POST',
+        url: context.POST_SURVEY_BOT_CHAT_URL,
+      },
+    });
+
+  const messageResult = await client.chat
+    .services(context.CHAT_SERVICE_SID)
+    .channels(channelSid)
+    .messages.create({
+      body: 'Hey! Before you leave, can you answer a few questions about this contact?',
```
nickhurlburt, line 82: Will this need to be localized in some way?

functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -0,0 +1,91 @@
...
+      const taskAttributes = JSON.parse(event.TaskAttributes!);
+
+      if (taskAttributes.isSurveyTask) {
+        await wait(3000); // wait 3 seconds just in case some bot message is pending
+
+        const handlerPath = Runtime.getFunctions()['helpers/postSurveyJanitor'].path;
+        const postSurveyJanitor = require(handlerPath).postSurveyJanitor as PostSurveyJanitor;
+        await postSurveyJanitor(context, { channelSid: taskAttributes.channelSid, channelType: 'chat' });
+  
+        const message = `Event ${TASK_CREATED_EVENT} handled by /helpers/postSurveyJanitor`;
```
nickhurlburt, line 72: I don't think you want `TASK_CREATED_EVENT` here.

functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -0,0 +1,91 @@
...
+const TASK_CANCELED_EVENT = 'task.canceled';
+const TASK_COMPLETED_EVENT = 'task.completed';
+
+const wait = (ms: number): Promise<void> => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+};
+
+export const handler = async (
```
nickhurlburt, line 34: This file could use some comments to give a quick summary of what we're doing with each event type.  This is only going to get bigger, and keeping a comment block of how we're routing events will be really important to understanding what's happening.

nickhurlburt (commented):
Some very impressive work here to make this feature work in the Twilio world.

There are several places here where naming is unclear and/or there's a lack of comments. 
 Having a separate document that describes post-survey will be very helpful to address that, but it would still be good to add key comments here.  It's very normal for the naming and comment issues to pop up in a very big PR like this that was created over time and made up of small PRs, but now it's worth a read through to clean up.

There is also some technical debt we are introducing due to duplicating (or almost duplicating) code from flex-plugins, especially for Insights. I'd like to know what you think we should do about this.  How long could this debt last until it needs to be addressed?  What are different options for addressing it?

There is also a heavy use of `console.log`, and we probably need to remove many of these before we take this to production.

I am not sure how quickly some of the things I mentioned, such as string localization, will need to be handled.  It might be that we just want this to get played with on development for awhile.  Then the console logs and localization issues aren't important.  But we should handle them before this gets to production.

functions/helpers/insightsService.private.ts
```diff
@@ -0,0 +1,87 @@
...
+  conversations?: { [key: string]: string | number };
+  customers?: { [key: string]: string | number };
+};
+
+type TaskAttributes = {} & InsightsAttributes;
+
+type InsightsObject = 'customers' | 'conversations';
+
+// Each of this ConfigSpec maps (possibly) many form field to one insights attribute
+export type OneToManyConfigSpec = {
```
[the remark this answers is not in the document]
GPaoloni: I agree that we should find a good name for this. If we change it, we should also do it in Flex. The purpose of an object  `OneToManyConfigSpec` is to describe in which task attribute field we should save one or more values, that's why I've called it one to many. The values are grabbed and then concatenated using a delimiter (`;`). There difference between the type in this repo vs the one in Flex is that here we use a property `questions` that's the list of question names to search for in the bot's memory, while in Flex we use a property `paths` that refers to the dot-joined paths where we want to grab values from the HRM forms.

functions/helpers/insightsService.private.ts
```diff
@@ -0,0 +1,87 @@
+import { get } from 'lodash';
```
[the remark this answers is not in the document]
GPaoloni: I think we should move all the definitions into a remote location as `json` files, and have services in both repos that given the definition version name fetches it from this "definitions library". S3 + Lambdas can be a nice fit for this. We can also bundle the definitions in Twilio Serverless as assets, but that will include all the definitions in all the environments. I don't see this as a big issue tho, but at some point we'll have 100 definitions and each helpline will use just a few of them. Another benefit we have from S3 is the privacy, as maybe we don't want to expose the data each helpline collects. If we do this we can either refactor a bit the definitions to isolate what can be changed without a new definition version (like layouts) from what forces a new version (like adding/renaming inputs from the forms), or we can move them as they are and refactor later on.

functions/postSurveyComplete.protected.ts
GPaoloni: Yes! I'm keeping the to debug, as they are very helpful, but they must be removed. Same goes for the console.logs in the Twitter integration files.

functions/postSurveyComplete.protected.ts
GPaoloni: Yes, it should be fetched from where we store our localized strings.

functions/postSurveyInit.ts
GPaoloni: Also yes.

functions/helpers/insightsService.private.ts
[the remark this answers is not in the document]
nickhurlburt: That sounds like a good approach for the `json` definition files. We will still need to do something about the code duplication -- the fact that `insightsService.ts` here and `insights Service.ts` in `flex-plugins` are almost the same thing. Some options (for the future):
1. Have a shared npm library that both import
2. Combine the code into one repo that has some common shared code but builds to multiple targets (eg, a flex-plugin and a serverless instance)

functions/postSurveyComplete.protected.ts
nickhurlburt: Can you add this comment?

nickhurlburt: Never mind... I see that it's at the top.

PR #97, opened 2021-09-20
Split deploy from health check actions

GPaoloni:
This PR is an attempt to split deployment from health check actions, in order to be one step closer to automating the entire setup process for a new helpline.

@nickhurlburt @murilovmachado I'm going to merge this against master to perform tests. In case something goes wrong it's easy to revert this merge commit. Nothing should be affected other than development deployments.


nickhurlburt:
@GPaoloni can you say more about why `.github/actions/healthCheck/dist/index.js` is checked in directly? I'm a little worried about a 9MB file being checked into our repo.

PR #103, opened 2021-09-20
Intentionally break healthCheck

GPaoloni:
This PR intentionally breaks https://github.com/techmatters/serverless/pull/102 to test if the workflow fails if health check fails. It will be reverted the tests.


GPaoloni:
Confirmed that if the health check fails the deployment action fails (tested by passing a wrong url).

Screenshot from 2021-09-20 18-20-59
Screenshot from 2021-09-20 18-21-14

PR #110, opened 2021-09-27
Health check rework

GPaoloni:


nickhurlburt (commented):
Could you add a short description so we know what it is when we look back at it?

PR #111, opened 2021-09-27
Added env vars required for case document feature

GPaoloni:
This PR adds the required env vars needed for case document to the deploy scripts of Aselo Dev.

Notes:
- We already have `secrets.AWS_DEFAULT_REGION` in GH secrets, do you think is really necessary to have `REGION` in Parameter Store?
- Same question goes for the AWS Aselo user's credentials, they could be hosted in GH secrets, but that will continue to split our sources of truth.
- I'm merging this PR to test it, but feel free to leave comments so I can address them.

### DIFF
--- a/.github/actions/deploy-beta/action.yml
+++ b/.github/actions/deploy-beta/action.yml
@@ -41,6 +41,18 @@ inputs:
   post-survey-bot-chat-url:
     description: 'Post Survey Bot Programmable Chat URL'
     required: true
+  aselo-app-access-key:
+    description: 'AWS credentials for Aselo user'
+    required: true
+  aselo-app-secret-key:
+    description: 'AWS credentials for Aselo user'
+    required: true
+  aws-region:
+    description: 'AWS region where Aselo documents bucket is hosted'
+    required: true
+  s3-bucket:
+    description: 'S3 bucket name where Aselo documents are stored'
+    required: true
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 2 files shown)


murilovmachado:
> * We already have `secrets.AWS_DEFAULT_REGION` in GH secrets, do you think is really necessary to have `REGION` in Parameter Store?

No, we only need one of them.

> * Same question goes for the AWS Aselo user's credentials, they could be hosted in GH secrets, but that will continue to split our sources of truth.

I think it would be good for us to define what will be stored as GitHub secrets, but in most scenarios I believe we should have those values on Prameter Store.

.github/actions/deploy-beta/action.yml
```diff
@@ -62,6 +74,10 @@ runs:
           HRM_STATIC_KEY=${{ inputs.hrm-static-key }}
           SURVEY_WORKFLOW_SID=${{ inputs.survey-workflow-sid }}
           POST_SURVEY_BOT_CHAT_URL=${{ inputs.post-survey-bot-chat-url }}
+          ASELO_APP_ACCESS_KEY=${{ inputs.aselo-app-access-key }}
+          ASELO_APP_SECRET_KEY=${{ inputs.aselo-app-secret-key }}
+          S3_BUCKET=${{ inputs.aws-region }}
+          AWS_REGION=${{ inputs.s3-bucket }}
```
murilovmachado, line 80: Out of these 4 values, there is only one actually changing from one configuration to another: `S3_BUCKET`. Maybe we don't need `ASELO_APP_ACCESS_KEY`, `ASELO_APP_SECRET_KEY` and `AWS_REGION` as inputs.

murilovmachado (commented):
Looks good, just one comment.

PR #112, opened 2021-09-28
Update staging deploy workflows

GPaoloni:
This PR updates all the deploy scripts for the staging tier accounts.

Notes: 
- Since here we are including values required for Post Survey and Case Document features, before deploying this we need to check that every resource mentioned in the workflows does exists (or create them if not).
- IMPORTANT! This deployments contain the breaking change from https://github.com/techmatters/serverless/pull/107.

### DIFF
--- a/.github/workflows/aselo_beta.yml
+++ b/.github/workflows/aselo_beta.yml
@@ -64,14 +64,49 @@ jobs:
         with:
           ssm_parameter: "PROD_TWILIO_AS_OPERATING_INFO_KEY"
           env_variable_name: "OPERATING_INFO_KEY"
-      - name: Set serverless base URL
+      - name: Set Twilio Flex Proxy service ID
         uses: "marvinpinto/action-inject-ssm-secrets@latest"
         with:
-          ssm_parameter: "ASELO_BETA_SERVERLESS_BASE_URL"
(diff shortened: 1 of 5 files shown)


.github/workflows/aselo_beta.yml
```diff
@@ -64,14 +64,49 @@ jobs:
...
-      - name: Set serverless base URL
+      - name: Set Twilio Flex Proxy service ID
         uses: "marvinpinto/action-inject-ssm-secrets@latest"
         with:
-          ssm_parameter: "ASELO_BETA_SERVERLESS_BASE_URL"
-          env_variable_name: "SERVERLESS_BASE_URL"          
-      # Call main-action to compile, deploy and check availability
-      - name: Executing main-action
-        uses: ./.github/actions/main-action
+          ssm_parameter: "PROD_TWILIO_AS_FLEX_PROXY_SERVICE_SID"
```
GPaoloni, line 70: Note that Aselo Beta is actually using `Production` environment. I think we can leave it as it is for now but it will be nice to change the variables on Parameter Store at some point.

nickhurlburt: Yes, it would be good to change that

murilovmachado (approved):
Looks good to me!

PR #113, opened 2021-09-28
Update production deploy workflows

GPaoloni:
This PR updates all the deploy scripts for the production tier accounts.

NOTE: since here we are including values required for Post Survey and Case Document features, before deploying this we need to check that every resource mentioned in the workflows does exists (or create them if not).


nickhurlburt (approved):
Looks good. I assume that once `deploy-beta` is our standard action, we will rename it to not be `beta`. Will we then need to do this again?

murilovmachado (approved):
Looks good!

PR #114, opened 2021-09-29
Fix Deploy Beta Actions

murilovmachado:
Primary reviewer: @GPaoloni 

This fixes an issue where `S3_BUCKET` was getting `AWS_REGION` value and vice versa.
Needs to redeploy every helpline that was deployed with this issue. Only development?


GPaoloni (approved):
Ups, my bad. Thanks for the fix!! No other helpline was deployed with this yet.

PR #139, opened 2021-11-03
Implemented slack messaging after a deployment

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-730

This PR updates `deploy-beta` action to send a Slack message after a deployment.


.github/actions/deploy-beta/action.yml
```diff
@@ -105,4 +111,13 @@ runs:
...
         TWILIO_AUTH_TOKEN: ${{ inputs.auth-token }}
-      shell: bash
\ No newline at end of file
+      shell: bash
+    # Send Slack notifying success
+    - name: Slack Aselo channel
+      id: slack
+      uses: slackapi/slack-github-action@v1.14.0
+      with:
+        channel-id: "C02L11XCKQU"
```
nickhurlburt, line 120: This doesn't seem all that sensitive, but would it be better as a GH secret?

nickhurlburt (commented):
Sweet!

.github/actions/deploy-beta/action.yml
GPaoloni: This is something I have a question about: this channel id is for the channel I've created. Should we use one of the existing ones or create a new one? Then I will move this to be a secret if preferred.

nickhurlburt: I think having an #aselo-deploys channel like you created is a great idea. Then it doesn't clutter other channels.

PR #146, opened 2021-11-30
Humairaa127 pr bt translations

Humairaa127:
Safernet localization changes:

- Created folder in assets/translations named pt-BR
- Added file to pt-BR: flexUI.json containing partial translations (pending helpline verification)
- Added file to pt-BR: messages.json containing chat intro and closing message


dee-luo:
@GPaoloni @murilovmachado Do we know if we need the CallerForm, CallerSection, and ContactDetails sections? I assumed this information would pull from the labels set in the form definitions (which are already translated) - or do they need to be explicitly translated here as well? 

@Humairaa127 do you have a list of what strings we left in English that we can check against once this gets deployed? We can do a secondary update on anything that was missed later.

nickhurlburt:
See comments in slack.

PR #158, opened 2021-12-06
Serverless deploy workflows: custom actions (VI)

GPaoloni:


.github/actions/custom-actions/action.yml
```diff
@@ -0,0 +1,7 @@
+name: 'Execute per-helpline custom action'
+runs:
+  using: "composite"
+  steps:
+    - name: Aselo Development release custom action
+      if: ${{ github.workflow }} == 'Aselo Development release'
```
nickhurlburt, line 6: Is this `if` statement temporary while we are testing this? Or will it be a big switch statement?  e.g.:

```
if: ${{ github.workflow }} == 'Aselo Development release'
uses: ./.github/actions/custom-actions/aselo_development_custom
if: ${{ github.workflow }} == 'Aselo Beta release'
uses: ./.github/actions/custom-actions/aselo_beta_custom
...
```

GPaoloni: No, this `if` is not temporary, and you are right, the idea is to put together a big switch-like statement (based on the workflow name we should have zero or exactly one case for every run). The reason is that we can't have dynamic statements in `uses` keyword (see https://github.com/actions/runner/issues/895), so we can't have something like `uses ./.github/actions/custom-actions/${{ github.workflow }}` (our original approach).

The only downside I can see of this approach is that we will need to perform 2 steps when adding a custom action:
- Write the custom action itself.
- Call that action if the workflow being executed is the target for the custom action one in `.github/actions/custom-actions/action.yml` (this file).

GPaoloni:
@nickhurlburt @murilovmachado  @stephenhand After merging this and https://github.com/techmatters/serverless/pull/166, seems like everything works as intended. To revert this we should first rever https://github.com/techmatters/serverless/pull/166. Let me know if you have any concerns. So far this only affects `Aselo Development release`. To make the rest of the workflows to use this, we should run `main-action` rather than `deploy-beta`.

PR #167, opened 2021-12-08
Remove pt-BR translations from serverless

nickhurlburt:
Primary reviewer: @GPaoloni 

This PR removes the `pt-BR` translations, which have moved to `flex-plugins` via https://github.com/techmatters/flex-plugins/pull/554. We want to have just one file for these translations to avoid confusion.


GPaoloni (approved):
Great, thanks!

PR #168, opened 2021-12-08
Enables post survey initial message to be translated

GPaoloni:
Jira: https://bugs.benetech.org/browse/CHI-829
This PR is related to https://github.com/techmatters/flex-plugins/pull/557

This PR:
- Adds `pt-BR` file for post survey translations. I think we will eventually add more messages to this, that's why I've created an isolated file, let me know if you prefer to have it merged with the already existing `messages.private.json` one.
- `postSurveyInit` will try to translate the initial message if there's a `taskLanguage` parameter being sent to the request.

### DIFF
--- a/functions/postSurveyInit.ts
+++ b/functions/postSurveyInit.ts
@@ -24,9 +24,13 @@ type EnvVars = {
 export type Body = {
   channelSid?: string;
   taskSid?: string;
+  taskLanguage?: string;
 };
 
-const createSurveyTask = async (context: Context<EnvVars>, event: Required<Body>) => {
+const createSurveyTask = async (
+  context: Context<EnvVars>,
+  event: Required<Omit<Body, 'taskLanguage'>>,
+) => {
   const client = context.getTwilioClient();
   const { channelSid, taskSid } = event;
 
@@ -60,7 +64,11 @@ const createSurveyTask = async (context: Context<EnvVars>, event: Required<Body>
   return surveyTask;
 };
 
-const triggerPostSurveyFlow = async (context: Context<EnvVars>, channelSid: string) => {
+const triggerPostSurveyFlow = async (
+  context: Context<EnvVars>,
+  channelSid: string,
+  message: string,
+) => {
   const client = context.getTwilioClient();
 
   await client.chat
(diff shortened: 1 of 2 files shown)


functions/postSurveyInit.ts
```diff
@@ -89,6 +94,27 @@ const triggerPostSurveyFlow = async (context: Context<EnvVars>, channelSid: stri
...
+  // Try to retrieve the triggerMessage for the approapriate language (if any)
+  const { taskLanguage } = event;
+  if (taskLanguage) {
+    try {
+      const translation = JSON.parse(
+        Runtime.getAssets()[`/translations/${taskLanguage}/postSurveyMessages.json`].open(),
+      );
+
+      if (!translation.triggerMessage)
+        console.error(`triggerMessage key is missing in translation for ${taskLanguage}`);
```
GPaoloni, line 107: Note that a few of this in production might trigger a page. Should I change to `warn` rather than `error` of is better to raise a lousy alarm if this is missing? (it will indicate some kind of missconfiguration taking place, but maybe not something to page someone).

stephenhand (approved):
Perhaps we should drop the 'console.log' warning on the linter check since console logging seems pretty usual in this project.

On logging, going forward we might consider adopting a logging framework like Winston in our repos. Even if we only ever log to stdout, it will allow us to add contextual information to every log statement, like timestamp, account, etc

But LGTM

GPaoloni:
> Perhaps we should drop the 'console.log' warning on the linter check since console logging seems pretty usual in this project.

Yeah we could do that.

> On logging, going forward we might consider adopting a logging framework like Winston in our repos. Even if we only ever log to stdout, it will allow us to add contextual information to every log statement, like timestamp, account, etc

I agree, using some framework can be helpful. In this particular repo we use lots of `console.warn` and `console.error`s cause Twilio triggers a webhook for each of them, that is then reported to our alerting system in AWS. If we go with something else, we should be test that the webhooks are still being called.

functions/postSurveyInit.ts
```diff
@@ -89,6 +94,27 @@ const triggerPostSurveyFlow = async (context: Context<EnvVars>, channelSid: stri
...
+
+      return translation.triggerMessage;
+    } catch {
+      console.error(`Couldn't retrieve triggerMessage translation for ${taskLanguage}`);
+    }
+  }
+
+  return 'Hey! Before you leave, can you answer a few questions about this contact?';
```
murilovmachado, line 115: This message could come from the assets en-US file, instead of hardcoded.

murilovmachado (approved):

functions/postSurveyInit.ts
GPaoloni: I think at some point is better to fallback to something in the code, just in case we can't access the assets and still be able to trigger the post survey. I will merge as it is for now but if you guys think this is better I will open a subsequent PR to address it.

PR #169, opened 2021-12-09
Updated Beta workflow to follow STG convention

GPaoloni:
This PR changes the secrets name for Aselo Beta account. Now Beta is considered the Staging environment, hence all of it's variables follows the convention pattern with `STG` env.

### DIFF
--- a/.github/workflows/aselo_beta.yml
+++ b/.github/workflows/aselo_beta.yml
@@ -108,8 +108,8 @@ jobs:
       - name: Executing deploy-beta
         uses: ./.github/actions/deploy-beta
         with:
-          account-sid: ${{ secrets.AS_BETA_ACCOUNT_SID }}
-          auth-token: ${{ secrets.AS_BETA_AUTH_TOKEN }}
+          account-sid: ${{ secrets.AS_PROD_ACCOUNT_SID }}
+          auth-token: ${{ secrets.AS_PROD_AUTH_TOKEN }}
           workspace-sid: $TWILIO_WORKSPACE_SID
           transfer-workflow-sid: $TWILIO_CHAT_TRANSFER_WORKFLOW_SID
           sync-service-api-key: $SYNC_SERVICE_API_KEY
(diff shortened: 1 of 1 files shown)


stephenhand (approved):

nickhurlburt:
This doesn't seem like a good idea to me. When we call something `PROD`, one of the implications is that it runs against our production backend, is accessed via our production VPN, and contains production data, especially PII. Beta does not do any of those things. While end users do see it, it's more similar to our staging accounts in that it's used for trying the product out without any guarantees of reliability and without special protections for PII.

murilovmachado:
> This doesn't seem like a good idea to me. When we call something `PROD`, one of the implications is that it runs against our production backend, is accessed via our production VPN, and contains production data, especially PII. Beta does not do any of those things. While end-users do see it, it's more similar to our staging accounts in that it's used for trying the product out without any guarantees of reliability and without special protections for PII.

I agree `PROD` doesn't seem like the correct **environment_shortname** name for Aselo Beta, but that's how it currently is everywhere on AWS (all resources for Aselo Beta are using PROD as environment short name).

This PR is changing only the git secrets names, so that it follows what's already being used for Aselo Beta: 
```
{ 
  shortName: 'AS',
  environmentShortName: 'PROD',
}
```

But I believe the correct configuration for Aselo Beta should be:
```
{ 
  shortName: 'AS',
  environmentShortName: 'STG', // Since it actually points to STG HRM
}
```
But I think this is out of scope for this PR. Here, Gian's only making sure that the git secret names follow a pre-existing pattern.

murilovmachado (approved):

nickhurlburt:
Changing over to use `STG` as an environment name sounds good to me. There's room for debate on whether the shortname should be `AS` or maybe `BE` or something. I suppose `AS` makes it consistent with Development's shortname, but I don't have a strong preference here.

nickhurlburt (approved):
Looks good!

PR #170, opened 2021-12-09
Added SaferNet workflow

stephenhand:
New prod workflow for SaferNet

### DIFF
--- a/.github/workflows/safernet_production.yml
+++ b/.github/workflows/safernet_production.yml
@@ -1,0 +1,127 @@
+# This is a basic workflow to publish a Twilio function with Github Actions
+
+name: SaferNet Brazil Production release
+
+# Controls when the action will run.
+on:
+  workflow_dispatch
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/safernet_production.yml
```diff
@@ -0,0 +1,127 @@
+# This is a basic workflow to publish a Twilio function with Github Actions
+
+name: SaferNet Brazil Production release
```
nickhurlburt, line 3: the staging one is 'SaferNet Staging release' without the Brazil... maybe we should do 'SaferNet Production release' to keep consistent with that

GPaoloni (approved):
Looks good!

.github/workflows/safernet_production.yml
```diff
@@ -0,0 +1,127 @@
...
+          env_variable_name: "FLEX_PROXY_SERVICE_SID"
+      - name: Set HRM static key to make HRM calls
+        uses: "marvinpinto/action-inject-ssm-secrets@latest"
+        with:
+          ssm_parameter: "PROD_TWILIO_SNBR_HRM_STATIC_KEY"
+          env_variable_name: "HRM_STATIC_KEY"
+      - name: Set Survey Workflow SID
+        uses: "marvinpinto/action-inject-ssm-secrets@latest"
+        with:
+          ssm_parameter: "PROD_TWILIO_SNBR_SURVEY_WORKFLOW_SID"
```
nickhurlburt, line 80: Is `SNBR` consistent with how we've been naming things? (If so, never mind me)

nickhurlburt (approved):
Rely on @GPaoloni's review more than mine. I just had some naming questions. Otherwise looks good.

PR #172, opened 2021-12-10
Change available check in assignOfflineContact

GPaoloni:
This PR fixes a potential bug when assigning an offline contact. Previously we looked for an available activity using the `friendlyName`, but now that we have helplines with different languages this might change, so the check is now changed to use the `available` property of the activities.

[:negative_squared_cross_mark:  ] Corresponding issue has been opened
[:heavy_check_mark: ] ~New tests added~ Tests matches this change

### Verification steps
- Start a local server with `npm run start`
- Expose the local server to the internet (using ngrok that can be done with `ngrok http <port>`, this requires configuring ngrok authtoken, you can obtain that login into https://dashboard.ngrok.com/get-started/setup).
- From Flex, change the Serverless url like `Twilio.Flex.Manager.getInstance().serviceConfiguration.attributes.serverless_base_url = '<your local server url>'`
- Submit an offline contact while being offline.

### DIFF
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -98,7 +98,7 @@ const assignToOfflineWorker = async (
   const availableActivity = await context
     .getTwilioClient()
     .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
-    .activities.list({ friendlyName: 'Available' });
+    .activities.list({ available: 'true' });
 
   await targetWorker.update({
     activitySid: availableActivity[0].sid,
--- a/functions/assignOfflineContact.ts
+++ b/functions/assignOfflineContact.ts
@@ -99,9 +99,12 @@ const assignToOfflineWorker = async (
     .getTwilioClient()
     .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
     .activities.list({ available: 'true' });
-if (availableActivity.length > 1) {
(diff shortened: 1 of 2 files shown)


nickhurlburt (approved):
Looks great!

functions/assignOfflineContact.ts
```diff
@@ -98,7 +98,7 @@ const assignToOfflineWorker = async (
   const availableActivity = await context
     .getTwilioClient()
     .taskrouter.workspaces(context.TWILIO_WORKSPACE_SID)
-    .activities.list({ friendlyName: 'Available' });
+    .activities.list({ available: 'true' });
```
nickhurlburt, line 102: One additional thing -- do we want to add a log if there are more than one of these, as discussed during our standup?  Such as:
```suggestion
if (availableActivity.length > 1) {
  console.warn(`There are ${availabilityActivity.length} available worker activities, but there should only be one.`);
}
```

functions/assignOfflineContact.ts
```diff
@@ -98,7 +98,12 @@ const assignToOfflineWorker = async (
...
+
+  if (availableActivity.length > 1) {
+    // eslint-disable-next-line no-console
+    console.warn(`There are ${availableActivity.length} available worker activities, but there should only be one.`);
```
murilovmachado, line 105: Nice!

murilovmachado (approved):
Looks good!

stephenhand (approved):

PR #173, opened 2021-12-13
Encode payload before signing it

murilovmachado:
## Description
Jira ticket: https://bugs.benetech.org/browse/CHI-924
Primary reviewer: @nickhurlburt 

This PR encodes the payload before generating the signature.
This solution was discussed with Caio from SaferNet and it fixes the issue we were getting with accents in the payload.

Note that we're only encoding the payload for creating the signature, not the body of the POST.

### DIFF
--- a/functions/saveContactToSaferNet.ts
+++ b/functions/saveContactToSaferNet.ts
@@ -34,13 +34,13 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
 
       const signedPayload = crypto
         .createHmac('sha256', SAFERNET_TOKEN)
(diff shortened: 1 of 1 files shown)


nickhurlburt (approved):

stephenhand (approved):

GPaoloni (approved):
Great! This way of generating `signedPayload` will continue to work for other helplines, right?

murilovmachado:
> Great! This way of generating `signedPayload` will continue to work for other helplines, right?

Yeah, it should work just fine.

PR #183, opened 2022-01-18
Update za.private.json

Humairaa127:
Update South Africa public holidays for 2022

### DIFF
--- a/assets/operatingInfo/za.private.json
+++ b/assets/operatingInfo/za.private.json
@@ -1,7 +1,7 @@
 {
   "timezone": "Africa/Johannesburg",
   "holidays": {
-    "01/02/2022":"New Years Day",
+    "01/02/2023":"New Years Day",
     "03/21/2022":"Human Rights Day",
     "04/15/2022":"Good Friday",
     "04/18/2022":"Family Day",
--- a/assets/operatingInfo/za.private.json
+++ b/assets/operatingInfo/za.private.json
@@ -1,18 +1,18 @@
 {
   "timezone": "Africa/Johannesburg",
   "holidays": {
-    "01/01/2021":"New Years Day",
-    "03/22/2021":"Human Rights Day",
-    "04/02/2021":"Good Friday",
-    "04/05/2021":"Family Day",
-    "04/27/2021":"Freedom Day",
-    "05/01/2021":"Workers Day",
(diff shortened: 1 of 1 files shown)


assets/operatingInfo/za.private.json
```diff
@@ -1,18 +1,18 @@
...
-    "04/05/2021":"Family Day",
-    "04/27/2021":"Freedom Day",
-    "05/01/2021":"Workers Day",
-    "06/16/2021":"Youth Day",
-    "08/09/2021":"National Womens Day",
-    "09/24/2021":"Heritage Day",
-    "12/16/2021":"Day of Reconciliation",
-    "12/25/2021":"Christmas",
-    "12/27/2021":"Boxing Day"
+    "01/02/2022":"New Years Day",
```
murilovmachado, line 4: Is the update New Year's day really Jan 2nd?

Humairaa127: > Is the updated New Year's day really Jan 2nd?

Thanks, yes for 2023.

murilovmachado (approved):

GPaoloni:
Looking the changes required here, I realize this is not ergonomic in the sense that it must be changed every year. We could improve the code to only consider day and month. Is that something desired @dee-luo @Humairaa127?

stephenhand:
> Looking the changes required here, I realize this is not ergonomic in the sense that it must be changed every year. We could improve the code to only consider day and month. Is that something desired @dee-luo @Humairaa127?

I'm guessing they need to be updated every year anyway, since public holiday day dates change every year (Easter, when the day of a regular holiday falls on the weekend they move the public holiday to the following Monday etc.)

I think the only way we could take the legwork out of this would be to import them from an external data source

Humairaa127:
> Looking the changes required here, I realize this is not ergonomic in the sense that it must be changed every year. We could improve the code to only consider day and month. Is that something desired @dee-luo @Humairaa127?

Yes, this is on the list of possible improvements i'm looking at for the year but not right now since it's a low effort process in it's current state, so we have time to work on it. As Steve mentioned, many dates change over years, and there are also certain countries who's holidays work on different calendars, not Gregorian. I've begun some basic research into libraries that are available for this and begin a team discussion soon to brainstorm solutions and assess viability.

PR #185, opened 2022-01-21
Implement Save Pending Contacts from Sync List

murilovmachado:
## Description
This PR creates a Serverless Function that tries saving the pending contacts and updates the pending-contacts sync list.

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://bugs.benetech.org/browse/CHI-880

### DIFF
--- a/functions/savePendingContacts.ts
+++ b/functions/savePendingContacts.ts
@@ -43,7 +43,7 @@ const getSaveContactFn = (
       // Callback passed to saveContactHandler
       const callback: ServerlessCallback = (error: any, callbackPayload: any) => {
         const isError = error || ![200, 204].includes(callbackPayload?.statusCode);
-        return isError ? rejectCallback(callbackPayload) : resolveCallback(callbackPayload);
+        return isError ? rejectCallback(error) : resolveCallback(callbackPayload);
       };
 
       return saveContactHandler(context, { ...event, payload }, callback);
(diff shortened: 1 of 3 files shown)


functions/savePendingContacts.ts
```diff
@@ -0,0 +1,127 @@
...
+  event: any,
+): Nullable<SaveContactFn> => {
+  const functionPath = Runtime.getFunctions()[functionName]?.path;
+  if (!functionPath) return;
+
+  const saveContactHandler = require(functionPath)?.handler;
+  if (!saveContactHandler) return;
+
+  // eslint-disable-next-line consistent-return
+  return payload =>
```
stephenhand, line 41: Would ```return async payload => {
  const response = await promisify(saveContactHandler)(context, { ...event, payload });
  if ([200, 204].includes(response.statusCode)) throw new Error(...);  
}
```
work?

File: functions/savePendingContacts.ts [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+    new Promise((resolveCallback, rejectCallback) => {
+      // Callback passed to saveContactHandler
+      const callback: ServerlessCallback = (error: any, callbackPayload: any) => {
+        const isError = error || ![200, 204].includes(callbackPayload?.statusCode);
+        return isError ? rejectCallback(callbackPayload) : resolveCallback(callbackPayload);
```
stephenhand, line 46: Would we not need to `rejectCallback(error)` if error was set?

functions/savePendingContacts.ts
```diff
@@ -0,0 +1,127 @@
...
+      const sharedStateClient = context.getTwilioClient().sync.services(SYNC_SERVICE_SID);
+      const list = await sharedStateClient.syncLists('pending-contacts').fetch();
+      const pendingContacts = await list.syncListItems().list();
+
+      type SyncListItemInstance = typeof pendingContacts[0];
+
+      const successfulRetriesListItems: SyncListItemInstance[] = [];
+      const failedRetriesListItems: SyncListItemInstance[] = [];
+
+      await Promise.all(
```
stephenhand, line 77: Might it be better to perform the whole save -> update / remove pending item operation in one parallelised operation (i.e. one `Promise.all(...)`)?

That might be more efficient that doing all the Save ops in parallel, waiting for them all to finish, then doing all the pending state changes. Also, if one of save ops faIls it won't interfere with the successful saves going on to update their state? Unless I'm missing something about the logic.

functions/savePendingContacts.ts
```diff
@@ -0,0 +1,127 @@
...
+        err.message &&
+        err.message.includes(
+          `The requested resource /Services/${SYNC_SERVICE_SID}/Lists/pending-contacts was not found`,
+        )
+      ) {
+        // Function should not fail or alarm if the sync list 'pending-contacts' doesn't exist yet
+        return resolve(success('The sync list pending-contacts was not found'));
+      }
+      // eslint-disable-next-line no-console
+      console.error(err);
```
stephenhand, line 123: Could this result in a double countedCloudfront error for the purposes of alarm triggers?

murilovmachado: I remember we had a discussion where we've spotted that `console.error()` were being counted as errors in Twilio. But I don't remember if those `console.error()` were from Twilio Functions or from somewhere else. Do you remember?

If that's the case, we'd need to change all Twilio Functions. They're all using `console.error()` at the catch block.

functions/savePendingContacts.ts
murilovmachado: Yes, the whole flux (saving => updating/deleting) can be done in parallel between each pending contact. When deleting, we use `ListItem.index`, but this `index` is not actually an index, but an id. So it can be done in parallel. I'll make the change.

functions/savePendingContacts.ts
stephenhand: I think that's what @GPaoloni was saying, we should remove the `console.error` or reduce the level in all our serverless functions because they double count, can you confirm Gian?

functions/savePendingContacts.ts
murilovmachado: Good catch! (although we're not using whatever comes back)

functions/savePendingContacts.ts
murilovmachado: Couldn't make it work like that

functions/savePendingContacts.ts
murilovmachado: I've changed it to **console.warn()** on this serverless function only. Later on, we might change it for all other serverless functions.

functions/savePendingContacts.ts
murilovmachado: Done

functions/savePendingContacts.ts
murilovmachado: Fixed

PR #187, opened 2022-01-28
Added support for Live_Report and Reporter_Country_ID from env vars

GPaoloni:
## Description
This PR adds options to provide "live" flag (to report "in production mode") and country id to the reports to IWF. This are optional environment variables in the Serverless instance, and if not provided, we'll just use `'T'` (test) flag and `null` country id.
This env vars should be added in the same custom action as the IWF credentials. Examples can be seen in the Aselo Development custom action (the code is commented out tho).

Instructions for the future selves can be found in https://benetech.app.box.com/notes/911942535061.


dee-luo (approved):
Set up doc LGTM!

stephenhand (approved):

PR #190, opened 2022-02-07
Factor out custom channel webhooks helpers

GPaoloni:
## Description
This PR
- Factors out common patterns used in the webhooks for custom channels.
- Reworks Twitter webhooks to use above changes.

Notes: 
- This PR is easier to review commit by commit.
- All `console.log`s will be removed before this is merged.

### Checklist
- [x] Corresponding issue has been opened. Kinda?
- [x] New tests added
- [N/A] Strings are localized

### Verification steps
Currently, Aselo dev webhook is configured to point to the -dev- instance of Twilio Serverless, so you can just exchange Twitter messages with @pgian5 to try this (already deployed to -dev- instance only).


functions/helpers/customChannels/customChannelToFlex.private.ts
```diff
@@ -0,0 +1,298 @@
...
+    channelSid: string;
+  },
+) =>
+  context
+    .getTwilioClient()
+    .chat.services(chatServiceSid)
+    .channels(channelSid)
+    .remove();
+
+export type AseloCustomChannels = 'twitter' | 'instagram';
```
stephenhand, line 106:
```suggestion
export enum AseloCustomChannel {
 Twitter = 'twitter',
 Instagram = 'instagram'
}
```

I know enums and literal unions do the same thing, but could we perhaps use enums where possible? It's just more obvious when you refer to an enum value in code that it's something predefined & from a restricted set of values, whereas a literal might just be a random string, or it might be from a restricted union, you can't immediately tell

functions/webhooks/twitter/FlexToTwitter.protected.ts
```diff
@@ -68,60 +67,57 @@ export const handler = async (
...
         return;
-      }
-
-      console.log('Message ignored (do not re-send self messages)');
-      resolve(success('Ignored event.'));
-      return;
+      case 'ignored':
+        resolve(success('Ignored event.'));
+        return;
+      default:
```
stephenhand, line 116: Should the default clause be necessary? I thought the types `redirectMessageToExternalChat` was permitted to return would ensure it can't be anything other than the 2 cases?

stephenhand (commented):
Couple of comments, looks good overall

functions/webhooks/twitter/FlexToTwitter.protected.ts
```diff
@@ -68,60 +67,57 @@ export const handler = async (
   const resolve = bindResolve(callback)(response);
 
   try {
-    const { recipientId, Body } = event;
-
+    const { recipientId, Body, From, ChannelSid, EventType, Source } = event;
     if (recipientId === undefined) {
```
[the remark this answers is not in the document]
GPaoloni: I think the only possible values here are string or undefined, since this function is actually [passed as a callback from our code](https://github.com/techmatters/serverless/blob/4a9df561844ff076758e76591a492db2d56590a2/functions/webhooks/twitter/TwitterToFlex.ts#L168) (and nowhere else it's being used).

Happy to consider adding more advance validation tho, is true that this could be "empty string" and in that case we'll end with a 5xx status code.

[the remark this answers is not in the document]
stephenhand: I was suggesting simpler validation, just `if (!recipientId) {` but it doesn't matter. I only brought it up because it caused some of the old tests to break but those are fixed now.

functions/helpers/customChannels/customChannelToFlex.private.ts
GPaoloni: Yeah I may be relying too much on the inference of VSCode and that's why using unions is clear to me (hovering on a type hints the union, when type is inferred autocompletion offers the possible values, etc). 
Happy to change to enum here, however I have read in several places ([like here](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums)) that enums are more of a use-case than the default to-go, so I'm not so convinced with the "where possible" part :P

GPaoloni: I've implemented the changes here https://github.com/techmatters/serverless/pull/206. Feel free to merge against the base PR if you think is good!

stephenhand (approved):

PR #200, opened 2022-02-14
Added post survey config for CA

GPaoloni:
## Description
Adds post survey config for the questions that are being tested in CA staging.


dee-luo (changes requested):
Can we just update the question to supportNeeds now? And make the corresponding change in Autopilot? 

We'll need to update the Insights attributes, as I don't think the ones listed here exist. Let's use: conversation_attribute_9, conversation_attribute_10, and customer_attribute_3.

Humairaa127 (approved):

PR #191, opened 2022-02-07
Added Instagram webhooks

GPaoloni:
## Description
This PR adds the required webhooks to integrate Instagram messages, both using the factored out helpers from the base branch.

~WIP:~
- ~Validate webhook signature.~
- ~Add tests.~
- ~All `console.log`s will be removed before this is merged.~

### Checklist
- [x] Corresponding issue has been opened. Kinda?
- [x] New tests added
- [N/A] Strings are localized

### Verification steps
Currently, Aselo dev webhook is configured to point to the -dev- instance of Twilio Serverless, so you can just exchange Instagram messages with @aselodevelopment to try this (already deployed to -dev- instance only). Not sure if this works with any IG account other than mine, because I'm set as adming of it and the app is not live but in development mode.

PR #204, opened 2022-02-16
Added support for injecting twitter related SSM parameter values into…

stephenhand:
Primary reviewer: @GPaoloni 

## Description
* Added support for injecting twitter related SSM parameter values into serverless deployments. 
* Added relevant twitter SSM parameters to ZM prod & staging workflows

TODO: Add Instagram variables

### Checklist
- [ X ] Corresponding issue has been opened
- [ N/A ] New tests added
- [ N/A ] Strings are localized

### Related Issues
Fixes CHI-1060

### Verification steps

Deploy ZM Staging
Check that the twitter environment variables are set correctly for zambia staging in the service in Twilio Console

### DIFF
--- a/.github/actions/custom-actions/aselo_development_custom/action.yml
+++ b/.github/actions/custom-actions/aselo_development_custom/action.yml
@@ -1,4 +1,5 @@
 name: 'Aselo Development release custom action'
+description: 'Sets up environment variables related to non-standard additional features in use in the development environment'
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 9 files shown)


GPaoloni (commented):
While this is not wrong and we could go with this approach, what I actually had in mind was to bundle the non-required credentials in custom actions rather than as part of the `main-action`. That way the files under `workflows` will look always the same, and the differing pieces of each helpline will be scoped under the [custom-actions](https://github.com/techmatters/serverless/tree/master/.github/actions/custom-actions).
There are already some of this in place, being [development custom action](https://github.com/techmatters/serverless/blob/master/.github/actions/custom-actions/aselo_development_custom/action.yml) the one I tend to use as example. 
Custom actions will only be executed if there is any for the current workflow being executed (https://github.com/techmatters/serverless/blob/4a9df561844ff076758e76591a492db2d56590a2/.github/actions/main-action/action.yml#L95), and this is achieved following the naming convention of the `.yaml` files.

.github/actions/custom-actions/aselo_development_custom/action.yml
```diff
@@ -28,20 +29,70 @@ runs:
...
+  - name: Set helpline Facebook Page Access Token
+    uses: "marvinpinto/action-inject-ssm-secrets@latest"
+    with:
+      ssm_parameter: "FACEBOOK_PAGE_ACCESS_TOKEN_105220114492262_TL-Staging"
+      env_variable_name: "FACEBOOK_PAGE_ACCESS_TOKEN"
+  - name: Set helpline Instagram Flex Flow SID
+    uses: "marvinpinto/action-inject-ssm-secrets@latest"
+    with:
+      ssm_parameter: "DEV_TWILIO_AS_INSTAGRAM_FLEX_FLOW_SID"
+      env_variable_name: "TWILIO_INSTAGRAM_FLEX_FLOW_SID"
```
GPaoloni, line 66: Because of the recent changes in https://github.com/techmatters/serverless/pull/191/commits/0e77b6dbe8ebfc714036fc5828003b1441cf47a8, we also need `FACEBOOK_APP_SECRET` from Parameter Store (same name in SSM). Can we add here before merging? (I can push the commit by myself if you want).

GPaoloni (approved):
Awesome, thanks!!

PR #217, opened 2022-02-24
Custom helpline workflow

stephenhand:
## Description

I want to try testing some service changes using a local version of flex. I know you can hack the code in flex to point at the local  HRM and use Aselo Dev for everything else, but I wanted to experiment with a fully isolated personal dev setup, using a separate Twilio account.

In order to do this I need to be able to deploy serverless to arbitrary accounts, so I've created a deployment workflow where you can specify the helpline & environment code yourself

It assumes all the SSM parameters are already set up (so you ideally need to have run a `terraform apply` on the target Twilio account) and github secrets have been populated

### Verification steps

* Try deploying the serverless functions to a correctly configured custom helpline, then use them via flex

### DIFF
--- a/.github/actions/custom-actions/aselo_development_custom/action.yml
+++ b/.github/actions/custom-actions/aselo_development_custom/action.yml
@@ -19,94 +19,94 @@ runs:
       with:
         ssm_parameter: "IWF_API_URL"
         env_variable_name: "IWF_API_URL"
-    # - name: Set IWF_API_ENVIRONMENT
-    #   uses: "marvinpinto/action-inject-ssm-secrets@latest"
-    #   with:
-    #     ssm_parameter: "DEV_AS_IWF_API_ENVIRONMENT"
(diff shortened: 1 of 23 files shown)


GPaoloni (approved):
The code looks good to me, but I fell it can lead to deployments in actual accounts that will be hard to track (is there a way to know the inputs for a given run?). I have 100% trust on you using it, but I'll be more confident if we can add some kind of warning/confirmation before the action is executed to make it clear: this is a dev tool and not an actual deploy action.

stephenhand:
> The code looks good to me, but I fell it can lead to deployments in actual accounts that will be hard to track (is there a way to know the inputs for a given run?). I have 100% trust on you using it, but I'll be more confident if we can add some kind of warning/confirmation before the action is executed to make it clear: this is a dev tool and not an actual deploy action.

I'll see if there's a way to make it more obvious it's only for dev use - even if it's just the naming, but I'll merge it now to test it works (since I suspect it won't, not without tweaking anyway)

PR #219, opened 2022-02-24
Fixed wrong usage of Twitter channel

GPaoloni:
## Description
Fixes a bug introduced in https://github.com/techmatters/serverless/pull/206: `instagram` `channelType` was being set for Twitter based tasks.
Also fixes a bad spelling.


GPaoloni:
@nickhurlburt @murilovmachado @stephenhand Merging this without review as I think is a trivial fix, and the bug is already deployed to Dev, I'm concerned any deployment to other environments will introduce this there too.

PR #227, opened 2022-03-14
Added CSAM credentials for ZM staging

GPaoloni:
This PR adds CSAM credentials for ZM Staging to be provided from custom action.


GPaoloni:
Merging without review as is a staging environment.

PR #229, opened 2022-03-15
Add static key auth to savePendingContacts

murilovmachado:
Primary reviewer: @GPaoloni 

## Description
This changes how some functions related to save pending contacts handle authorization.

- `/savePendingContacts`: Previously used `twilio-flex-token-validator.functionValidator` wrapper. Now it looks for a static key.
- `/saveContactToSaferNet`: Previously used `twilio-flex-token-validator.functionValidator` wrapper. Now it uses a combination of `twilio-flex-token-validator.validator` (notice it's not `functionValidator`) or checks for a static key.

With these changes we're able to achieve two things:

1. An external function (AWS lambda) can call `/savePendingContacts` in a secure way using a static key
2. `/saveContactToSaferNet` can be called in a secure way from two different sources: from Aselo app using a Twilio's generated AccessToken; or from an external function using a static key.

### Related Issues
Fixes https://bugs.benetech.org/browse/CHI-1065
lambda PR: https://github.com/techmatters/infrastructure-config

### DIFF
--- a/.github/actions/custom-actions/safernet_production_custom/action.yml
+++ b/.github/actions/custom-actions/safernet_production_custom/action.yml
@@ -13,7 +13,7 @@ runs:
       with:
         ssm_parameter: "PROD_SAFERNET_TOKEN"
         env_variable_name: "SAFERNET_TOKEN"
-    - name: Set SAFERNET_ENDPOINT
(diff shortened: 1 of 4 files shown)


stephenhand (approved):
Looks good, once I understand it :-)

PR #230, opened 2022-03-16
Added country code and LIVE flag

GPaoloni:
## Description
Set environment flag to live ("L") and country code to ZM contry code (can be found in https://bugs.benetech.org/secure/RapidBoard.jspa?rapidView=19&view=detail&selectedIssue=CHI-1089).


GPaoloni:
Merging without review as I need to test this changes.

PR #231, opened 2022-03-16
Fix saveContactToSaferNet to accept objects or string

murilovmachado:
## Description
Small bug correction:
When `/saveContactToSaferNet` is called via lambda, a payload object was being passed instead of a payload string.
This PR fixes `/saveContactToSaferNet` so that it works with either payload type.


murilovmachado:
Merging this fix so `/saveContactToSaferNet` will work properly when called from the lambda function. But feel free to review, it's a very small fix.

PR #237, opened 2022-04-01
Fix filepath for insights post-survey config

dee-luo:
I think the filepath for the Canada post survey insights file was wrong - no data has been saving in Insights. I'm fixing the filepath to match what we have in Dev (and what's in the post-survey set up instructions).


GPaoloni (approved):
Great!

PR #250, opened 2022-06-07
adding twitter custom action for beta environment

janorivera:
##Reviewer
@GPaoloni 

## Description

Adding the custom action for twitter env variables for the beta environment


.github/actions/custom-actions/aselo_staging_custom/action.yml
```diff
@@ -0,0 +1,48 @@
+name: 'Aselo Beta release custom action'
+description: 'Adds Twitter related environment variables to the Aselo Beta environment'
```
GPaoloni: I think the description should be something more general. Is true that _for now_ we use this for Twitter creds only, but if we ever want Instagram in Beta, this same action will be extended.

GPaoloni (commented):
This is good, but it also needs to be referenced in the [main of the custom actions](https://github.com/techmatters/serverless/blob/master/.github/actions/custom-actions/action.yml), so it will be executed when the proper workflow (the Beta one) is executed.

GPaoloni (approved):

PR #251, opened 2022-06-08
Support Instagram message deletion webhooks

GPaoloni:

## Description
This PR adds support for Instagram message deletion according to the specs in the ticket (link below). The changes are:
- Fixed wrong Parameter Store credential being pulled from in Aselo Development deployment (Facebook page name has changed and credentials regenerated so the old one is not valid anymore).
- Exporting helper function to retrieve the active chat channel for a given user (if it exists). If no channel is returned, then the user does not have an active conversation.
- Added support to include message attributes when sending a message to Flex in a custom channel.
- All Instagram messages now include it's message id (coming from Instagram) in the message attributes, under the propery `messageExternalId`.
- When a message is unset (deleted) on Instagram:
  - If the user has an active chat, looks for the message in the chat channel that matches with the `messageExternalId`, and edit it to indicate it has been deleted.
  - If the user does not have active channel, we skip the event.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1174)
- [x] New tests added

### Verification steps
Not sure anyone else can test this, as only people with developer role in the IG app can trigger the webhooks. In case you can, this modifications are already deployed to Aselo dev environment, so to test:
- Send an Instagram message to @aselodevelopment ig account.
- Receive the resulting task in Flex.
- [Optional] Exchange some messages.
- Delete one or more messages sent by the user and confirm they are properly deleted.
- Capture the chat channel sid to use in later step.
- End the chat.
- Try deleting messages now that there is no active conversation (nothing should happen).
- Check that the deleting messages on Instagram does not delete any message on Twilio system when chat is not active anymore.

### DIFF
--- a/functions/webhooks/instagram/InstagramToFlex.ts
+++ b/functions/webhooks/instagram/InstagramToFlex.ts
@@ -32,6 +32,7 @@ type InstagramMessageObject = {
     mid: string;
     text?: string; // the body of the message
     attachments?: { type: string; payload: { url: string } }[];
+    is_deleted?: boolean;
   };
 };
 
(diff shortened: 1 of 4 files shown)


functions/webhooks/instagram/InstagramToFlex.ts
```diff
@@ -62,6 +63,29 @@ const shouldFilterMessage = (message: InstagramMessageObject['message']) => {
...
+    messageExternalId,
+  }: { chatServiceSid: string; channelSid: string; messageExternalId: string },
+) => {
+  const client = context.getTwilioClient();
+  const messages = await client.chat
+    .services(chatServiceSid)
+    .channels(channelSid)
+    .messages.list();
+
+  const messageToUnsed = messages.find(
```
stephenhand, line 80: Typo ` messageToUnsend`

functions/webhooks/instagram/InstagramToFlex.ts
```diff
@@ -62,6 +63,29 @@ const shouldFilterMessage = (message: InstagramMessageObject['message']) => {
...
+    m => JSON.parse(m.attributes).messageExternalId === messageExternalId,
+  );
+
+  const unsent = await messageToUnsed?.update({ body: 'The user has unsent this message' });
```
stephenhand, line 84: Will this need translating?

stephenhand (approved):

PR #253, opened 2022-06-10
Support Instagram story mentions

GPaoloni:

## Description
This PR handles story tagging in a slightly different way. Before we were always ignoring this events, now we:
- Ignore the event if the user does not have an active covnersation.
- Redirect the media cdn (temporary url) of the story if the user has an active conversation.
Screenshot from 2022-06-10 17-45-38

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1173)
- [x] New tests added

### Verification steps
Not sure anyone else can test this, as only people with developer role in the IG app can trigger the webhooks. In case you can, this modifications are already deployed to Aselo dev environment, so to test:
- Send an Instagram message to @aselodevelopment ig account.
- Receive the resulting task in Flex.
- [Optional] Exchange some messages.
- Tag in a story the  @aselodevelopment and see the message is properly displayed as above image shows.
- Capture the chat channel sid to use in later step.
- End the chat.
- Tag in a story the  @aselodevelopment now that there is no active conversation (nothing should happen).
- Check that no message was sent on Twilio system when chat is not active anymore.

### DIFF
--- a/functions/webhooks/instagram/InstagramToFlex.ts
+++ b/functions/webhooks/instagram/InstagramToFlex.ts
@@ -56,12 +56,15 @@ export type Body = InstagramMessageEvent & {
   bodyAsString?: string; // entire payload as string (preserves the ordering to decode and compare with xHubSignature)
 };
 
-const shouldFilterMessage = (message: InstagramMessageObject['message']) => {
-  // Filter story mention
-  if (message.attachments && message.attachments[0].type === 'story_mention') return true;
+const isMessageDeleted = (message: InstagramMessageObject['message']) => message.is_deleted;
 
-  return false;
-};
+const isStoryMention = (message: InstagramMessageObject['message']) =>
+  message.attachments && message.attachments[0].type === 'story_mention';
+
+const fromStoryMention = (message: InstagramMessageObject['message']) =>
+  message.attachments
+    ? `Story mention: ${message.attachments[0].payload.url}`
+    : 'Looks like this event does not incluydes a valid url in the payload';
 
 const unsendMessage = async (
   context: Context,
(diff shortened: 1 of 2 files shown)


functions/webhooks/instagram/InstagramToFlex.ts
```diff
@@ -56,12 +56,15 @@ export type Body = InstagramMessageEvent & {
...
 
-  return false;
-};
+const isStoryMention = (message: InstagramMessageObject['message']) =>
+  message.attachments && message.attachments[0].type === 'story_mention';
+
+const fromStoryMention = (message: InstagramMessageObject['message']) =>
+  message.attachments
+    ? `Story mention: ${message.attachments[0].payload.url}`
+    : 'Looks like this event does not incluydes a valid url in the payload';
```
murilovmachado, line 67: typo at `incluydes `

functions/webhooks/instagram/InstagramToFlex.ts
```diff
@@ -56,12 +56,15 @@ export type Body = InstagramMessageEvent & {
...
-  // Filter story mention
-  if (message.attachments && message.attachments[0].type === 'story_mention') return true;
+const isMessageDeleted = (message: InstagramMessageObject['message']) => message.is_deleted;
 
-  return false;
-};
+const isStoryMention = (message: InstagramMessageObject['message']) =>
+  message.attachments && message.attachments[0].type === 'story_mention';
+
+const fromStoryMention = (message: InstagramMessageObject['message']) =>
```
murilovmachado, line 64: I was a little bit confused with this function being named `fromStoryMention`. It was not clear what it would return. Maybe `getStoryMentionText()` is a more readable name?

murilovmachado (approved):
I've suggested some very minor things, but it already looks good to me.
(I haven't run the code, just reviewed the PR)

stephenhand (approved):

PR #255, opened 2022-06-17
Added IWF vars to IN prod (test mode)

GPaoloni:
## Description
Added required env vars to IN prod for IWF reporting.

NOTE: as `IWF_API_ENVIRONMENT` variable is commented out in the action, test mode will be deployed. Once we want this to go live we must create `PROD_IN_IWF_API_ENVIRONMENT` Parameter Store (AWS) variable with the `L` value and un-comment the corresponding code in `.github/actions/custom-actions/aarambh_production_custom/action.yml`

### DIFF
--- a/.github/actions/custom-actions/aarambh_production_custom/action.yml
+++ b/.github/actions/custom-actions/aarambh_production_custom/action.yml
@@ -1,0 +1,48 @@
+name: 'Aarambh Production release custom action'
+description: 'Sets up environment variables related to non-standard additional features in use in the Aarambh Production environment'
+runs:
+  using: "composite"
+  steps:
+    # Set IWF credentials
+    - name: Set IWF_API_USERNAME
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: "IWF_API_USERNAME"
+        env_variable_name: "IWF_API_USERNAME"
+    - name: Set IWF_API_PASSWORD
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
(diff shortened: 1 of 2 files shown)


nickhurlburt (commented):
This looks good to me too, but since it's production we should have at least one other person look at it.

.github/actions/custom-actions/aarambh_production_custom/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: "IWF_API_PASSWORD"
+        env_variable_name: "IWF_API_PASSWORD"
+    - name: Set IWF_API_URL
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: "IWF_API_URL"
+        env_variable_name: "IWF_API_URL"
+    # - name: Set IWF_API_ENVIRONMENT
```
stephenhand, line 22: Why are these commented out? Are they not required? I've been adding them but we should delete them from all the workflows that set up CSAM if they aren't required, to stop idiots like me copying them everywhere :-)

stephenhand (approved):

.github/actions/custom-actions/aarambh_production_custom/action.yml
GPaoloni: They default to "test" mode and null country code. For now I left that commented out intentionally but will change prior to merge this, once @Humairaa127 confirms that the same setup is working fine for staging environment.

PR #256, opened 2022-06-17
Added IWF vars to IN stg

GPaoloni:
## Description
Added required env vars to IN staging for IWF reporting.

NOTE: this does not includes country code nor environment as we only want for testing mode prior to enabling in production account.


nickhurlburt (commented):
Looks good to me!

PR #257, opened 2022-06-17
Debug flag

stephenhand:


nickhurlburt:
@stephenhand should this stay open? Should it be reviewed?

PR #263, opened 2022-07-04
Adding twitter custom action for CO STG

janorivera:
## Description
Modifying the "main" action to include the TeGuio Staging Custom action and adding the custom action  teguio_staging for the twitter setup.

### DIFF
--- a/.github/actions/custom-actions/action.yml
+++ b/.github/actions/custom-actions/action.yml
@@ -24,6 +24,9 @@ runs:
     - name: Safernet Production release custom action
       if: github.workflow == format('SaferNet Production release')
       uses: ./.github/actions/custom-actions/safernet_production_custom
+    - name: TeGuio Staging release custom action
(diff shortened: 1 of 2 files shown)


GPaoloni:
> maybe we need to standardise the naming of the TWITTER_ACCESS_TOKEN and TWITTER_ACCESS_TOKEN_SECRET SSM parameters? They seem a bit all over the place rn? Something like <short_helpline><short_environment> ?

@stephenhand The naming was chosen having in mind the case where a single Twilio account might have more than one Twitter (or Instagram or whatever) account associated. Currently the Serverless functions are not capable of taking care of that scenario, but the changes would be rather small. There was a [PoC](https://github.com/techmatters/serverless/pull/75/files) for this a while ago. I'm happy to change the naming as needed if that's more convenient, just pointing out why I choose to go with the current thing :P

PR #264, opened 2022-07-06
Updating to match existing working workflow

janorivera:
I think that `github.workflow == format('Te Guio Colombia II Staging release') `should match an existing working workflow, it wasn't the case before so I updated it.

@acedeywin @stephenhand  also maybe is worth deleting the previous te guio workflow?


GPaoloni (approved):
Approving cause I know this will allow you to continue working, but we probably want to schedule some time to get rid of the broken one an name the working one accordingly.

PR #265, opened 2022-07-08
CHI-1309: Custom channel ids

stephenhand:
Includes changes for #255 - perhaps wait for this to be merged to review

Primary Reviewer @GPaoloni 

## Description

Allows custom channel IDs to be set in payloads we report to IWF, as they requested

### Checklist
- [X] New tests added

### Related Issues
Fixes CHI-1309

### Verification steps

?

### DIFF
--- a/functions/reportToIWF.ts
+++ b/functions/reportToIWF.ts
@@ -21,13 +21,14 @@ type EnvVars = {
   IWF_API_URL: string;
   IWF_API_ENVIRONMENT?: string;
   IWF_API_COUNTRY_CODE?: string;
+  IWF_API_CHANNEL_ID?: string;
 };
 
-type IWFReportPayload = {
+export type IWFReportPayload = {
   Reporting_Type: 'R'; // "R" for report
   Live_Report: 'L' | 'T'; // "L" for Live, "T" for test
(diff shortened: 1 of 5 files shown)


functions/reportToIWF.ts
```diff
@@ -72,11 +73,13 @@ export const handler: ServerlessFunctionSignature<EnvVars, Event> = TokenValidat
         ? parseInt(context.IWF_API_COUNTRY_CODE, 10)
         : null;
 
+      const channelID = context.IWF_API_CHANNEL_ID ? parseInt(context.IWF_API_CHANNEL_ID, 10) : 51;
```
GPaoloni, line 76: Is it fine that we still default to `51` if there's no custom id? Or is that a required field from now on? Just asking in case the requirements of IWF changed.

GPaoloni (approved):
Just a question that you have already made to the product team. Otherwise this looks great to me, thanks!

functions/reportToIWF.ts
stephenhand: From what I read in the ticket, they were specifically asking for those 3 accounts to be changed - I'm going to assume 51 is still a valid default until told otherwise

PR #269, opened 2022-07-15
Asset cleanup - Remove non private assets

mythilytm:
@GPaoloni 

## Description
- Remove aselo-dev asset as it is not required, and could possibly be causing deployment problems.


GPaoloni (approved):
Thanks!

For the record and for context to the rest of the team, since we had two assets with same name but one "public" and the other "private", we were reaching an error like
Screenshot from 2022-07-15 17-16-20

We resolved just keeping the private one.

PR #278, opened 2022-07-25
[Bug Fix] Custom channel user-to-channel maps linked to stale channels

GPaoloni:
Primary Reviewer: @murilovmachado 

## Description
This PR:
- Renames `postSurveyJanitor` to a more general and descriptive one: `chatChannelJanitor`.
- Exposes a small helper to indicate if a channel is `AseloCustomChannel`.
- Adds a new condition on taskrouter callback to perform cleanup on custom channels (prevents stale state).

Notes:
- After this PR is merged, check that the taskrouter webhook url points to `-production` (currently pointing to `-dev`).
- In order for this changes to take effect in existing accounts, we need to enable two more events on taskrouter webhook: `task.deleted` and `task.system-deleted`.
- Above item means that we need to address this in the terraform scripts too (deferring PR on that until this one is approved & merged).

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1321)
- [ ] New tests added

#### To test cancelled event:
- Decrease the timeout in "send to flex" widget for Twitter or Instagram, under Messaging Studio Flow (both works on the same functions) to something that allows for faster iterations (30 seconds should be enough).
- Check that your user does not have a channel-to-user map under sync -> documents.
- Send a new message via the selected platform.
- DO NOT accepted the generated task in Flex, we want it to expire (so it's cancelled).
- Go in and check that now you do have a channel-to-user map under sync -> documents. Copy the channel sid in it.
- After the task is cancelled, check that the channel-to-user map under sync -> documents is deleted. Also the channel used (above copeid sid) should be marked as INACTIVE.
- :exclamation: Remember to rollback any changes you've made in the Studio Flow.
#### To test deleted event:
- Decrease the "send to flex" widget for Twitter or Instagram (both works on the same functions) to something that allows for faster iterations.
- Check that your user does not have a channel-to-user map under sync -> documents.
- Send a new message via the selected platform.
- Accept the generated task in Flex.
- Go in and check that now you do have a channel-to-user map under sync -> documents. Copy the channel sid in it.
- Go to taskrouter -> tasks and deleted the task you got generated from the Twilio console.
- After the task is deleted, check that the channel-to-user map under sync -> documents is deleted. Also the channel used (above copeid sid) should be marked as INACTIVE.

### DIFF
--- a/functions/helpers/chatChannelJanitor.private.ts
+++ b/functions/helpers/chatChannelJanitor.private.ts
@@ -56,26 +56,30 @@ const deactivateChannel = async (
 
   const attributes = JSON.parse(channel.attributes);
   
-  if (attributes.proxySession) {
-    await deleteProxySession(context, attributes.proxySession);
-  }
+  if (attributes.status !== 'INACTIVE') {
+    if (attributes.proxySession) {
+      await deleteProxySession(context, attributes.proxySession);
+    }
+    
+    const newAttributes = { ...attributes, status: 'INACTIVE' };
+    const updated = await channel.update({
+      attributes: JSON.stringify(newAttributes),
+      xTwilioWebhookEnabled: 'true',
+    });
   
-  const newAttributes = { ...attributes, status: 'INACTIVE' };
-  const updated = await channel.update({
-    attributes: JSON.stringify(newAttributes),
-    xTwilioWebhookEnabled: 'true',
-  });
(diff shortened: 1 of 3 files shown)


murilovmachado (approved):
Code looks good to me, Gian!

I didn't go through the verification steps, but I'm happy with you merging this.

PR #289, opened 2022-08-17
Reduce error logs

GPaoloni:
Primary reviewer: @nickhurlburt 

## Description
This PR attempts to reduce the amount of times we get paged:
- Removes all `console.error`s in `catch` blocks that already raise a Twilio error webhook (as the function will return status code 500).
- Adds error handling on `functions/helpers/addCustomerExternalId.private.ts`, instead than throwing we'll `console.error`, so the Twilio error webhook is called once per occurrence. If this still occurs that can indicate an internal error.
- Moves around check to avoid calling above function for offline contacts, that is where we found conflicts because of the race condition between "removing task that can't be assigned" in `assignOfflineContact` and "adding external id" in the taskrouter callback. Can rollback if we prefer to keep that check inside of it tho.

### Checklist
- [x] New tests added

### Verification steps
- I can help with this after deploying to development :)

### DIFF
--- a/functions/helpers/addCustomerExternalId.private.ts
+++ b/functions/helpers/addCustomerExternalId.private.ts
@@ -1,5 +1,6 @@
 // eslint-disable-next-line prettier/prettier
 import type { Context } from '@twilio-labs/serverless-runtime-types/types';
+import type { TaskInstance } from 'twilio/lib/rest/taskrouter/v1/workspace/task';
 
 export type Body = {
   EventType: string;
@@ -11,6 +12,12 @@ type EnvVars = {
 };
 
 const TASK_CREATED_EVENT = 'task.created';
+
+const logAndReturnError = (taskSid: TaskInstance['sid'], workspaceSid: EnvVars['TWILIO_WORKSPACE_SID'], step: 'fetch' | 'update', errorInstance: unknown) => {
+  const errorMessage = `Error at addCustomerExternalId: task with sid ${taskSid} does not exists in workspace ${workspaceSid} when trying to ${step} it.`;
+  console.error(errorMessage, errorInstance);
+  return { message: errorMessage };
+};
 
 export const addCustomerExternalId = async (context: Context<EnvVars>, event: Body) => {
   console.log('-------- addCustomerExternalId execution --------');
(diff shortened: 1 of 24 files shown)


nickhurlburt:
As an additional comment, removing the `console.error`s means that we will stop getting reported errors like:
image

But continue to get errors like:
image

As far as I can tell, we still get all the same information in the resulting log, so having just one of these seems fine, and is less chatty.

functions/helpers/addCustomerExternalId.private.ts
```diff
@@ -12,6 +13,12 @@ type EnvVars = {
 
 const TASK_CREATED_EVENT = 'task.created';
 
+const logAndReturnError = (taskSid: TaskInstance['sid'], workspaceSid: EnvVars['TWILIO_WORKSPACE_SID'], step: 'fetch' | 'update', errorInstance: unknown) => {
+  const errorMessage = `Error at addCustomerExternalId: task with sid ${taskSid} does not exists in workspace ${workspaceSid} when trying to ${step} it.`;
+  console.error(errorMessage, errorInstance);
```
nickhurlburt, line 18: Will this create the duplicate error messages that we are getting rid of elsewhere? Or are we returning a 200 in Line 19, so it's still only one error in the logs?

This does feel like a much more clean and clear message than we had before.

nickhurlburt (approved):
One question, but otherwise good to go!  Thank you!

functions/helpers/addCustomerExternalId.private.ts
GPaoloni: This will catch the error in the fetch/update task steps, logging the error so the error reporter is raised once, but we'll still return an error message to `taskRouterCallback`, instead than causing it to throw. Hence, `taskRouterCallback` will just complete it's execution (not throwing), but the error will be reported and will appear in the monitor once because of this `console.error`.
The only place that could cause `taskRouterCallback` to really throw, is if the task's attributes object can't be parsed, which I don't think is even possible, but we can also wrap it in a try/catch if that makes sense.

nickhurlburt: Sounds good, thank you! I agree that checking for parseable attributes doesn't seem necessary.

PR #296, opened 2022-08-25
Chi 1359 update serverless npm dependencies

acedeywin:
## Description

- Update dependencies

### Checklist
- [N/A] New tests added

### Related Issues
Fixes Chi-1359


nickhurlburt:
Scanning over this, it seems like there are a lot of small reformatting changes. What is the cause of those? I just want to be sure that these are in line with the standard linting and formatting that everyone is using, so we don't end up changing the formatting every time a different person works with the code.

acedeywin:
@nickhurlburt  yes, they are in line with the standard linting and formatting we are using.

tests/assignOfflineContact.test.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
+// @ts-nocheck
```
GPaoloni, line 2: Same

tests/createContactlessTask.test.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
+// @ts-nocheck
```
GPaoloni, line 2: Same

tests/createContactlessTask.test.ts
```diff
@@ -101,7 +111,8 @@ describe('createContactlessTask', () => {
     };
 
     const { getTwilioClient, DOMAIN_NAME } = baseContext;
-    await createContactlessTask({ getTwilioClient, DOMAIN_NAME }, event1, callback1);
+    const payload: any = { getTwilioClient, DOMAIN_NAME };
+    await createContactlessTask(payload, event1, callback1);
```
GPaoloni, line 115: Was this refactor needed? Logically seems like doing the same to me.

tests/transferChatResolve.test.ts
```diff
@@ -143,7 +152,7 @@ describe('transferChatResolve', () => {
     };
 
     await Promise.all(
-      [{}, event1, event2, event3, event4].map(event =>
+      [event1, event2, event3, event4].map((event) =>
```
GPaoloni, line 155: This is removing one test case?

GPaoloni (commented):
A few comments on things that I would do differently, but there might be reasons why you are doing this way. Open to debate (and worth asking opinion for the rest of the team).

Definitely we don't want to "shut down eslint/TS" for sure :P

functions/adjustChatCapacity.ts
```diff
@@ -85,12 +86,12 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
       if (workerSid === undefined) return resolve(error400('workerSid'));
       if (adjustment === undefined) return resolve(error400('adjustment'));
 
-      const validBody = { workerSid, adjustment };
+      const validBody = { workerSid, adjustment, request: { cookies: {}, headers: {} } };
```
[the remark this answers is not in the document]
acedeywin: @GPaoloni removing `request` will break the code. It's a required update that was made in @twilio-labs and it was added as a type under `ServerlessFunctionSignature`. So in ordeer for the tests and updates to work, I had to add it to most part of the code beacuse it is required.

tests/adjustChatCapacity.test.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
+// @ts-nocheck
```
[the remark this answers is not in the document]
acedeywin: @GPaoloni Oh, this were not supposed to be here. I was testing solutions on how to solve the over 71 linting problems that I was having :)

.eslintrc.js
```diff
@@ -18,5 +18,9 @@ module.exports = {
     // Place to specify ESLint rules. Can be used to overwrite rules specified from the extended configs
     // e.g. "@typescript-eslint/explicit-function-return-type": "off",
     "import/prefer-default-export": "off",
+    "@typescript-eslint/camelcase": "off",
+    "@typescript-eslint/indent": "off",
+    "no-console": "off",
+    "no-shadow": "off"
```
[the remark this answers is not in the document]
acedeywin: @GPaoloni I was having serious linting issues as a result of the updates and "disable next line" was not working rather, it was compounding to the problems. So I spoke to Steve and he advised that I make use of the rules within the `.eslintrc.js` config file. So the solution was to turn off those linters as seen in the `.eslintrc.js` file. It doesn't disable eslint in these files but just turned off the above error. It was literally what "disable next line" was supposed to solve but for reasons which I wasn't sure of, that wasn't the case.

tests/transferChatResolve.test.ts
acedeywin: @GPaoloni , I will refactor it. The problem here is passing an empty `{}` will throw an error due to Typescript is expecting `request` within the empty object. I will refactor it to be so.

tests/createContactlessTask.test.ts
acedeywin: @GPaoloni  Yes, the refactor was needed. createContactlessTask was expecting three parameters and not an object as part of the parameter. So Typescript wasn't allowing it to fly. I had to refactor the object to a single variable.

functions/adjustChatCapacity.ts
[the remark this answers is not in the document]
GPaoloni: @acedeywin Right. I did not got TS errors but didn't tried to compile, only used the TS linter of VSCode. Anyway I think this comment is actually for this comment https://github.com/techmatters/serverless/pull/296#discussion_r958565914. I still think we should not carry the entire `Body` object into functions that do not require it, that's why I suggested only picking the ones relevant to the internal function.

functions/reportToIWF.ts
```diff
@@ -75,7 +89,25 @@ export const handler: ServerlessFunctionSignature<EnvVars, Event> = TokenValidat
...
+          | 'Reported_Category_ID'
+          | 'Reported_URL'
+          | 'Reporter_Anonymous'
+          | 'Reporter_First_Name'
+          | 'Reporter_Last_Name'
+          | 'Reporter_Email_ID'
+          | 'Reporter_Country_ID'
+          | 'Reporter_Description'
+        >
+      > = {
```
GPaoloni, line 110: I don't think you need to define the type again here, since this is picking all properties from `IWFReportPayload`.

`body: IWFReportPayload = { .... `

functions/reportToIWF.ts
```diff
@@ -47,11 +47,25 @@ export type Event = {
...
+        | 'Reported_URL'
+        | 'Reporter_Anonymous'
+        | 'Reporter_Description'
+        | 'Reporter_Email_ID'
+        | 'Reporter_First_Name'
+        | 'Reporter_Last_Name'
+      >
+    >,
+    callback: ServerlessCallback,
+  ) => {
```
GPaoloni, line 68: This change does not seems correct. Why modify the `Event` type at all if we don't pass that information to the `handler`? `event: Event` might be the one you want here?

GPaoloni (approved):
Two small changes but otherwise looks good

functions/adjustChatCapacity.ts
```diff
@@ -90,7 +91,9 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
       const { status, message } = await adjustChatCapacity(context, validBody);
 
       return resolve(send(status)({ message, status }));
-    } catch (err) {
+    } catch (err: any) {
+      // eslint-disable-next-line no-console
+      console.error(err);
```
GPaoloni, line 96: This is reintroducing stuff we don't want in

functions/assignOfflineContact.ts
```diff
@@ -214,7 +236,9 @@ export const handler: ServerlessFunctionSignature = TokenValidator(
       }
 
       resolve(success(assignmentResult.newTask));
-    } catch (err) {
+    } catch (err: any) {
+      // eslint-disable-next-line no-console
+      console.error(err);
```
GPaoloni, line 241: Same

GPaoloni (commented):
@acedeywin Sorry to bother but there are a lot of `console.error` statements being re.introduced in the code that we don't want. Probably they slipped in when you resolved the merge conflicts against master? I'm not sure tbh

functions/adjustChatCapacity.ts
GPaoloni: Yes but we get rid of most of them [here](https://github.com/techmatters/serverless/pull/289). Let's keep them out please.

functions/webhooks/line/FlexToLine.protected.ts
```diff
@@ -27,32 +27,30 @@ export type Body = Partial<WebhookEvent> & {
   recipientId?: string; // The Line id of the user that started the conversation. Provided as query parameter
 };
 
-const sendLineMessage = (context: Context<EnvVars>) => async (
```
GPaoloni, line 30: @murilovmachado Can you confirm if this is the final version, or if this is outdated?

functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -53,14 +53,24 @@ const wait = (ms: number): Promise<void> => {
   });
 };
 
-const isCreateContactTask = (eventType: EventType, taskAttributes: { isContactlessTask?: boolean }) => 
-  eventType === TASK_CREATED_EVENT && taskAttributes.isContactlessTask;
+const isCreateContactTask = (
+  eventType: EventType,
+  taskAttributes: { isContactlessTask?: boolean },
+) => eventType === TASK_CREATED_EVENT && taskAttributes.isContactlessTask;
```
GPaoloni, line 59: Be aware that this was updated, you will need to fix the merge conflict, and please keep the version from `master`.

GPaoloni (approved):

functions/webhooks/line/FlexToLine.protected.ts
murilovmachado: @GPaoloni @acedeywin  Strange that this PR deletes this line. This function should exist.

functions/webhooks/taskrouterCallback.protected.ts
acedeywin: @GPaoloni Alright.

package.json
```diff
@@ -16,48 +16,47 @@
...
+    "twilio-run": "^3.4.3",
+    "typescript": "^4.7.4"
   },
   "engines": {
     "node": "^12.0"
   },
   "dependencies": {
-    "@tech-matters/serverless-helpers": "^2.0.0",
-    "@twilio-labs/serverless-runtime-types": "^1.1.8",
-    "@twilio/runtime-handler": "1.2.3",
```
murilovmachado, line 51: @GPaoloni @acedeywin This is removing `@twilio/runtime-handler` so it breaks some stuff on our serverless functions

murilovmachado: Looking at Twilio Serverless, this change made it downgrade to version `1.0.1` instead of `1.2.3`, and this breaks some stuff. Can you update this dependency or rollback this change?

acedeywin: @murilovmachado @GPaoloni  It has been fixed. Can you check to see if anything still breaks?

PR #305, opened 2022-09-05
[Bug Fix] Fixed isCreateContactTask condition

GPaoloni:
## Description
I realized I introduced a wrong condition [here](https://github.com/techmatters/serverless/pull/289), because if `taskAttributes.isContactlessTask` is true then that is an offline contact. The correct predicate should negate this condition as in this PR.

This bug is currently deployed to JM prod so this should be prioritized!


mythilytm (approved):
Looks good!

PR #313, opened 2022-09-12
Create action.yml

Humairaa127:
## Description
Add Line custom actions file for thailand staging


murilovmachado (changes requested):
This file looks good, there's one more file that you'll need to edit so this code is run:
`.github/actions/custom-actions/action.yml`. You'll need to add something like:

```
- name: Childline Thailand Staging release custom action
      if: github.workflow == format('Thailand Staging release')
      uses: ./.github/actions/custom-actions/thailand_staging_custom
```

PR #315, opened 2022-09-13
Wrapper functionValidator for twilio-flex-token-validator

GPaoloni:
Primary reviewer: @stephenhand 

## Description
This PR bundles in a wrapper around https://github.com/twilio/twilio-flex-token-validator. This wrapper is bundled as part of our `tech-matters-serverless-helpers-lib` package, to make easier it's usage. A single function is exposed, `functionValidator`, that have the exact behavior of [Twilio's `functionValidator`](https://github.com/twilio/twilio-flex-token-validator#using-in-twilio-function), except that it also accepts a second argument, `options: { allowGuestToken?: boolean } = {}`, which can be used to protect a function from public access but allow guest tokens to call it (probably relevant for @mythilytm's current work). Two commits inside of this branch exhibit this wrapper working as expected: [test-functionValidator (reject guests)](https://github.com/techmatters/serverless/pull/315/commits/c064355cce4eefc2e8c096e4bc1c4c191d0143f3) and [test-functionValidator (accept guests)](https://github.com/techmatters/serverless/pull/315/commits/547288d5cd1a84f182d5c9e80fbb448f11529d50). You can sit on those commits to test locally, but the test function is not part of this PR.

Other changes in this PR:
- I'm bundling `tech-matters-serverless-helpers-lib` package locally. I will test if this works in the actual Twilio runtime or not, and if it does we can decide to preserve it local or if we want to continue publishing the package to npm.
- Mocks the `functionValidator` in every function that uses it, since it is now returning 403 in all the test files for functions that are wrapped with it. I ignore the actual details on how but looks like Twilio's one was "doing nothing" in test executions.

WIP:
- [x] Test if locally installed lib works in Twilio runtime. EDIT: [no they don't](https://github.com/techmatters/serverless/actions/runs/3049253316/jobs/4915165114) :P
- [x] Deploy updated `tech-matters-serverless-helpers-lib`, and revert local bundle.
- [x] [Add tests](https://github.com/techmatters/serverless/blob/gian_token-validator-wrapper/tech-matters-serverless-helpers-lib/tests/tokenValidator.test.ts). No need to run separate CI steps for this folder while it lives inside of this repo, they are executed as part of the Serverless one. From https://github.com/techmatters/serverless/actions/runs/3055275957/jobs/4928150690:
  Screenshot from 2022-09-14 16-16-16

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1410)
- [x] New tests added

### Verification steps
- `git checkout c064355` to be at [test-functionValidator (reject guests)](https://github.com/techmatters/serverless/pull/315/commits/c064355cce4eefc2e8c096e4bc1c4c191d0143f3).
- `npm ci` to install this updates.
- `npm start` to start development server.
- Try calling `test-functionValidator` with different tokens:
  - curl -X GET 'http://localhost:3000/test-functionValidator?Token=' should return "Unauthorized: token was not provided."
  - curl -X GET 'http://localhost:3000/test-functionValidator?Token=<worker token>' should return "This is allowed :D, welcome in!"
  - curl -X GET 'http://localhost:3000/test-functionValidator?Token=<guest token>' should return "Unauthorized: endpoint not open to guest tokens."
- `git checkout 547288d` to be at [test-functionValidator (accept guests)](https://github.com/techmatters/serverless/pull/315/commits/547288d5cd1a84f182d5c9e80fbb448f11529d50).
- Repeat the steps for the previous branch, the only change should be that the guest token now is also accepted.


tech-matters-serverless-helpers-lib/src/tokenValidator.ts
```diff
@@ -0,0 +1,63 @@
...
+      return failedResponse(
+        'Unauthorized: AccountSid or AuthToken was not provided. For more information, please visit https://twilio.com/console/runtime/functions/configure',
+      );
+    }
+
+    if (!token) {
+      return failedResponse('Unauthorized: token was not provided.');
+    }
+
+    return validator(token, accountSid, authToken)
```
stephenhand, line 48: Just wondering why you decided to use the promise API directly rather than `async / await` here? Seems like `async / await` might be easier to read?

tech-matters-serverless-helpers-lib/tests/helpers.ts
```diff
@@ -0,0 +1,48 @@
+const Twilio = require('twilio');
```
stephenhand, line 1: Can we name this file something a little more descriptive? Like `twilioGlobals.ts` or something? 

Sorry to nitpick - modules with generic names like `helper.ts` or `util.ts` are my pet hate :-P

stephenhand (approved):
LGTM, just a couple of comments.

tech-matters-serverless-helpers-lib/tests/helpers.ts
GPaoloni: That makes sense, thanks for the feedback :P https://github.com/techmatters/serverless/pull/315/commits/0fb99be6cfed7e03b4fb1c6e47d12757ecab99f9

tech-matters-serverless-helpers-lib/src/tokenValidator.ts
GPaoloni: Changed to the async/await API in https://github.com/techmatters/serverless/pull/315/commits/0c0b68fab788d314c46488d97f529dab53f17ce9.

PR #316, opened 2022-09-21
Adding Twitter and Instagram to TH

janorivera:
## Description
Adding Twitter and Instagram env variables to TH Staging


GPaoloni (approved):
I suggest grouping the variables either like
- Set env (grab the values from SSM) for all channels
- Append (add to .env) for all channels

or

- Set and append channel 1
- Set and append channel 2
etc. 

Just for the sake of readability :P

PR #317, opened 2022-09-23
Chi 1259 workflow to deploy serverless helpline staging remove re-run

acedeywin:
## Description

- New feature: Deploy Serverless to all Helpline Staging Environments

### Checklist
- [N/A] New tests added

### Related Issues
Fixes CHI-1259

### Verification steps

This PR is an implementation for deploying Serverless to all Helpline Staging Environments. The final solution works well, except that at least one helpline will fail in every deployment. I have written the different methods and hacks tried to re-run failed jobs. You can find them [here](https://benetech.app.box.com/notes/1022372979138).

So in every deployment, for now, the failed deployment will have to be run manually until we find a solution for an automatic re-run.


.github/workflows/aarambh_staging.yml
```diff
@@ -4,7 +4,12 @@ name: Aarambh Staging release
 
 # Controls when the action will run.
 on:
-  workflow_dispatch
```
stephenhand, line 7: We still need the 'workflow_dispatch' trigger so we can deploy the helplines individually if required

GPaoloni:
I also have a question that may be silly at this point :stuck_out_tongue:.
Have you tested that the custom actions are executed? Because those are including a lot of env vars for particular accounts, and the logic is based on the workflow being executed. If a reusable workflow loses it's "name" we lose them. Hopefully that's not the case.

stephenhand:
> I also have a question that may be silly at this point 😛. Have you tested that the custom actions are executed? Because those are including a lot of env vars for particular accounts, and the logic is based on the workflow being executed. If a reusable workflow loses it's "name" we lose them. Hopefully that's not the case.

Sorry, maybe I'm misunderstanding what you're saying. AFAIK, if the workflow names go out of sync with the all staging workflow, we will get a failing job for that deployment, rather than it just running the core action?

.github/workflows/aarambh_staging.yml
acedeywin: @stephenhand I am not sure the `workflow_dispatch` initially does anything. It was a stand-alone with nothing (like inputs) to trigger. It kept on breaking the code until I removed it.

.github/workflows/aarambh_staging.yml
```diff
@@ -106,6 +111,7 @@ jobs:
           env_variable_name: "S3_BUCKET"
       # Call main-action to compile and deploy
       - name: Executing main-action
+        if: always()
```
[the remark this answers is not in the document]
acedeywin: @GPaoloni yeah, I was supposed to remove that.

GPaoloni:
@stephenhand What I meant is that the custom actions are executed based on the workflow name. My question is "if `South Africa`  deploy workflow is called using this action, the custom actions get the workflow name `South Africa` or `Deploy All` "? I assume that the first, but asked just to confirm.

.github/workflows/aarambh_staging.yml
stephenhand: Without that we can't run the builds independently from the Github GUI (I tested this to confirm), so removing it is not an option. 

It's valid to have a `workflow_dispatch` without any options, you just have to pick a branch or tag to run it from, you need to fix whatever issues you were having that caused you to remove it initially

stephenhand:
> @stephenhand What I meant is that the custom actions are executed based on the workflow name. My question is "if `SafeSpot` deploy workflow is called using this action, the custom actions get the workflow name `SafeSpot` (and [this](https://github.com/techmatters/serverless/blob/015f55b066eb8ec3812711e801dc26b1a6e3f5de/.github/actions/custom-actions/action.yml#L22) is executed) or `Deploy All` "? I assume that the first, but asked just to confirm.

I see what you mean - yes that would have to be tested. We can always replace it with a passed in variable based on the short helpline & environment codes if it doesn't work - that might be better anyway, it will prevent people breaking things by renaming a workflow, which I think people might expect to be something they can do safely.

stephenhand (commented):
See comments (not sure why it didn't update my review status)

.github/workflows/aarambh_staging.yml
acedeywin: @stephenhand Fixed!

stephenhand (approved):
LGTM, but as @GPaoloni pointed out, we need to specifically test that the custom actions are performed for each helpline (before we merge), if we haven't done already. 

You should be able to verify these actions are being run from the log output of the workflow

janorivera:
hey @acedeywin  could you also add the new Telefon Zaufania (PL) Staging flow into this?

GPaoloni:
@acedeywin I have put together a [small PR](https://github.com/techmatters/serverless/pull/326) that shows how we can use the approach Steve suggested here, that is to me the easiest to go with. Of course this is up to debate, but the same changes that are working for development in that PR should work for the staging ones you have here?

acedeywin:
@GPaoloni @stephenhand Everything seems to be working fine. Custom actions get deployed efficiently. Thanks @GPaoloni

.github/workflows/aarambh_staging.yml
```diff
@@ -98,12 +105,12 @@ jobs:
         uses: "marvinpinto/action-inject-ssm-secrets@latest"
         with:
           ssm_parameter: "REGION"
-          env_variable_name: "AWS_REGION"
+          env_variable_name: "AWS_REGION"
```
stephenhand, line 108: Can we revert the whitespace changes?

.github/workflows/custom_helpline.yml
```diff
@@ -15,14 +15,26 @@ on:
...
     env:
-      TWILIO_SSM_PREFIX: ${{github.event.inputs.environment_code}}_TWILIO_${{github.event.inputs.helpline_code}}
-      TWILIO_ACCOUNT_SID_SECRET_NAME: ${{github.event.inputs.helpline_code}}_${{github.event.inputs.environment_code}}_ACCOUNT_SID
-      TWILIO_AUTH_TOKEN_SECRET_NAME: ${{github.event.inputs.helpline_code}}_${{github.event.inputs.environment_code}}_AUTH_TOKEN
+      TWILIO_SSM_PREFIX: ${{inputs.environment_code}}_TWILIO_${{inputs.helpline_code}}
+      TWILIO_ACCOUNT_SID_SECRET_NAME: ${{inputs.helpline_code}}_${{inputs.environment_code}}_ACCOUNT_SID
+      TWILIO_AUTH_TOKEN_SECRET_NAME: ${{inputs.helpline_code}}_${{inputs.environment_code}}_AUTH_TOKEN
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js
```
stephenhand, line 40: Can we add the helpline-name parameter to the custom workflow too? We can create the value from `${{inputs.helpline_code}}_${{inputs.environment_code}}`

stephenhand (approved):
Couple of comments but looks good overall

PR #319, opened 2022-10-03
Taskrouter Callback to trigger postsurvey on task wrapup

mythilytm:
Primary reviewer: @GPaoloni 

## Description
- Taskrouter Callback is modified to trigger post survey on task wrapup instead of on task complete
- In Twilio Console for Flex Task Assignment, taskRouterCallback webhook has been updated to trigger on task wrapup (This will have to be manually updated for current helpline accounts in Twilio, and for future account set up, terraform module for task router logic)

### Checklist
- [n/a] New tests added

### Related Issues
Fixes #[1409](https://bugs.benetech.org/browse/CHI-1409)

### Verification steps
- This can be verified by following steps in this PR in Flex - https://github.com/techmatters/flex-plugins/pull/937

### DIFF
--- a/functions/webhooks/taskrouterCallback.protected.ts
+++ b/functions/webhooks/taskrouterCallback.protected.ts
@@ -61,8 +61,7 @@ const isCreateContactTask = (
 ) => eventType === TASK_CREATED_EVENT && !taskAttributes.isContactlessTask;
 
 const isCleanupPostSurvey = (eventType: EventType, taskAttributes: { isSurveyTask?: boolean }) =>
-  (eventType === TASK_CANCELED_EVENT || eventType === TASK_WRAPUP) &&
-  taskAttributes.isSurveyTask;
+  (eventType === TASK_CANCELED_EVENT || eventType === TASK_WRAPUP) && taskAttributes.isSurveyTask;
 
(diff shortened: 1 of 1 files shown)


GPaoloni (approved):
LGTM!

>In Twilio Console for Flex Task Assignment, taskRouterCallback webhook has been updated to trigger on task wrapup (This will have to be manually updated for current helpline accounts in Twilio, and for future account set up, terraform module for task router logic)

I suggest emphatically communicating this to all the team members, via Slack and if possible adding to a "next deploy plan TODOs". 
Also it would be awesome if you could make the appropriate changes / create a Jira card so we don't forget doing so. 
If I'm correct, we need to enable `task.wrapup` AND disable `task.completed` (since no other logic here is making use of it)?

GPaoloni:
@mythilytm Sorry, I missunderstood the intent of this PR. I don't think that this changes are really necessary! Since the task being cleaned on `isCleanupPostSurvey` is not the actual contact task, but the post survey task, I believe we still want to preserve that logic in the complete event. Is there a particular reason why move this to `wrapup` instead? Remember that if `isPostSurvey` is `true`, that task is not the actual "contact in Flex" that a counselor works on, but the task generated for the `Survey` task queue so we keep track of the survey results in isnights.

mythilytm:
Closing this PR given the reasons mentioned by Gian.

PR #322, opened 2022-10-06
CHI-1441: Set playwright test run to use one worker when cutting a release

stephenhand:
## Description

Set E2E tests run on QA release to use 1 worker


GPaoloni (approved):
Thanks!

PR #326, opened 2022-10-11
[PoC]:Added logic to pass down helpline-identifier for custom actions (AS_DEV)

GPaoloni:
Primary reviewer: @acedeywin 

## Description
This PoC shows how to change our custom actions to depend in an input instead than the "workflow name" as we were doing.
- Adds `helpline-identifier` as a required parameter for `.github/actions/custom-actions/action.yml`, to decide if a custom step should be triggered.
- Adds `helpline-identifier` as a required parameter for `.github/actions/main-action/action.yml`, and passes it down to `custom-actions` step.
- Adds "AS_DEV" as the the `helpline-identifier` for the Aselo Development deployment workflow

As you can see in [this action execution](https://github.com/techmatters/serverless/actions/runs/3228170666/jobs/5283910469), the custom action is executed, and the expected steps are taking place (see [aselo development custom](https://github.com/techmatters/serverless/blob/master/.github/actions/custom-actions/aselo_development_custom/action.yml) for reference).
Screenshot from 2022-10-11 12-41-51

Next steps:
- If this `helpline-identifier` become required, that means that we have to change **every** workflow, whether it uses custom action or not. I think for sanity that makes sense, since we wanted the workflows to look all the exact same but using different helpline identifiers. This means that we can continue to use scripts to generate the workflows, and the only changes for custom actions will be within those files and not in the workflows. Can be convinced other way if you think this is not right.
- While this is a PoC, above item means we'll want to merge it if we decide that this is the easiest replacement for the "github workflow name" approach (which I do).

Asking review from @stephenhand that's already aware of this and suggested we take this path.

### Verification steps
- Run 
  ```curl -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer <yout GH token>" "https://api.github.com/repos/techmatters/serverless/actions/workflows/aselo_development.yml/dispatches" -d '{"ref":"gian_poc-custom-actions-helpline-name"}```
- Confirm that I'm not lying in the above statements :P

### DIFF
--- a/.github/actions/custom-actions/action.yml
+++ b/.github/actions/custom-actions/action.yml
@@ -1,5 +1,10 @@
 name: 'Execute per-helpline custom action'
 description: Compound 'routing' action to load custom deployment actions for the helpline being deployed
+inputs:
+  helpline-identifier:
+    description: 'The identifier in the format "<short helpline>-<short env>"" used for this helpline (e.g. "AS_DEV")'
+    required: true
+    type: string
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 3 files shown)


stephenhand:
Looks good to me.

I wish there was something a bit less clunky than the 'custom action picker action' to implement this logic but if there is I can't think of it

stephenhand (approved):

PR #323, opened 2022-10-07
End Chat function for webchat widget and update to taskRouterCallback

mythilytm:
## Description
- Updated taskRouterCallback at task creation stage to store taskSid in channelAttributes
- Defined endChat function to use channelSid and corresponding taskSid to change the assignmentStatus to 'wrapping' from initial 'assigned' (or 'canceled' if initial status is 'reserved').

### Checklist
- [x] Corresponding issue has been opened
- [ ] New tests added

### Related Issues
Fixes #[1402](https://bugs.benetech.org/projects/CHI/issues/CHI-1402) partially. See [PR ](https://github.com/techmatters/webchat/pull/126)

### Verification steps
- Ensure that this version of code is deployed to production

PR #328, opened 2022-10-11
Revert "End Chat function for webchat widget and update to taskRouterCallback"

mythilytm:
Reverts techmatters/serverless#323


nickhurlburt (approved):
This is fine. Note that for a revert, especially to remove a bug, no approval is necessary (though you can tag people so they are aware).

PR #330, opened 2022-10-13
Transfers Listener

murilovmachado:
Primary reviewer: @GPaoloni 

## Description
The goal of this PR is to move some transfers-related code that was being executed at the Counselors' browser into Serveless. This also fixes the reject transfer scenario, where the task was never going back to the original counselor.

Implementation details:
- `TaskrouterCallback` function now fetches a list of listeners
  - They are all inside `taskrouterListeners` folder
  - They all export a default object of `TaskrouterListener` type
  - Right now, the only listener is `TransfersListener`
  - The plan is to move all hardcoded pieces of tasrouterCallback into separate listeners
- For each listener
  - `TaskrouterCallback` checks if the listener `shouldHandle` that `event`
  - `TaskrouterCallback` calls the listener `handleEvent` passing down `context` and `event`
- `TransferChatStart` do not change the original task `assignmentStatus` to `wrapping`
  - It's kept `assigned`, so the counselor can have the control back in case the target rejects the transfer
  - It also doesn't kick the original from the channel while the call is not yet accepted/rejected
- `TaskrouterListener` basically handles 3 scenarios:
  - Accept chat transfer
  - Reject chat transfer
  - A cleanup specific for voice cold transfers
    - where Twilio updates the original reservation to `wrapup`
- `@tech-matters/serverless-helpers` now exposes a second module: `taskrouter`, that contains
  - All taskrouter event types
  - A definition type for the taskrouter event
  - A definition type for the taskrouter listeners: `TaskrouterListener`

There'll be a follow up front-end PR to remove old code and fix minor front-end stuff: https://github.com/techmatters/flex-plugins/pull/978

### Deployment
1. Enable the following taskrouterCallback events on Twilio console:
   1. Reservation Accepted
   2. Reservation Rejected
   3. Reservation Timeout
   4. Reservation Wrapup
2. Deploy Serverless
3. Deploy flex-plugins

If Serverless is deployed and flex-plugins is not, we will have issues around some controls visibility: such as `transfer` or `end chat` buttons will be visible/not visible at incorrect times. Not sure if we're going to have bigger issues such as some tasks not getting cleaned up properly.

 
### Checklist
- [x] New tests added

### Related Issues
Fixes https://bugs.benetech.org/browse/CHI-1350

### Verification steps
It needs to be deployed in order to test. After that, test accept/reject scenarios for chat/voice transfers.


functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -78,17 +66,42 @@ const isCleanupCustomChannel = (eventType: EventType, taskAttributes: { channelT
   return channelToFlex.isAseloCustomChannel(taskAttributes.channelType);
 };
 
+/**
+ * Fetch all taskrouter listeners from the listeners folder
+ */
+const getListeners = () => {
+  const functionsMap = Runtime.getFunctions();
```
stephenhand, line 74: We could definitely make this logic a little cleaner if we merge the shared module stuff, we could have a `register` method here and all the `TaskRouterListeners` register themselves with the `taskRouterCallback`. But what you have here is definitely the best approach within the current limitations

functions/webhooks/taskrouterCallback.protected.ts
```diff
@@ -79,15 +67,41 @@ const isCleanupCustomChannel = (eventType: EventType, taskAttributes: { channelT
...
+  event: EventFields,
   callback: ServerlessCallback,
 ) => {
   const response = responseWithCors();
   const resolve = bindResolve(callback)(response);
 
   try {
+    await runTaskrouterListeners(context, event, callback);
+
     const { EventType: eventType } = event;
```
stephenhand, line 105: Perhaps we can refactor this existing logic into TaskRouterListeners as well? Perhaps for another PR...

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -0,0 +1,225 @@
+/* eslint-disable global-require */
```
stephenhand, line 1: Should we have some tests for this?

stephenhand (approved):

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -0,0 +1,225 @@
...
+  TaskrouterListener,
+  EventFields,
+  EventType,
+  RESERVATION_ACCEPTED,
+  RESERVATION_REJECTED,
+  RESERVATION_TIMEOUT,
+  RESERVATION_WRAPUP,
+} from '@tech-matters/serverless-helpers/taskrouter';
+
+export const eventTypes: EventType[] = [
```
GPaoloni, line 16: This are the ones we listen to, right?

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -0,0 +1,225 @@
...
+        .workspaces(context.TWILIO_WORKSPACE_SID)
+        .tasks(originalTaskSid)
+        .update({
+          attributes: JSON.stringify(attributesWithChannelSid),
+        });
+
+      await client.taskrouter.workspaces(context.TWILIO_WORKSPACE_SID).tasks(taskSid).update({
+        assignmentStatus: 'canceled',
+        reason: 'task transferred rejected',
+      });
```
GPaoloni, line 158: Does it makes sense to run this two in parallel?

GPaoloni (approved):
Looks very good! Code so clean and well commented, really enjoyed this review haha. Awesome PR, thanks Mu!!

functions/webhooks/taskrouterCallback.protected.ts
murilovmachado: Yeah, the plan to do that is described at the PR description

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    console.log(`===== Executing TransfersListener for event: ${eventType} =====`);
+
+    const taskAttributes = JSON.parse(event.TaskAttributes!);
+
+    /**
+     * If a chat transfer gets accepted, it should:
+     * 1) Complete the original task
+     */
+    if (isChatTransferAccepted(eventType, taskChannelUniqueName, taskAttributes)) {
```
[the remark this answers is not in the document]
murilovmachado: Yes, we can early return

functions/taskrouterListeners/transfersListener.private.ts
murilovmachado: Nice idea!

PR #332, opened 2022-10-15
[POC ] tech-matters-serverless-helpers-lib as local module

stephenhand:
Experiment with local repos in serverless.

The problem with local repos in NPM is that unless you include them in a specific way, they can tend to take 'short cuts' (literally!), compared to remote packages, like symlinking from the package repo into the consuming project's node_modules rather than copying the relevant files

To fully simulate a local package, and thus work around the limitation imposed by the Twilio Serverless deploy process, you can put your local package into a `.tgz` file and import it from there, which this PR demonstrates

This PR is a POC with some changes that probably aren't required for a bare minimum solution to use a local module - such as upgrading to node 16 and moving the functions and the library into sibling modules, but are probably worthwhile changes to make in the long run.

Most of the challenges I faced were actually around getting it to build & the tests to pass when moving to a local module rather than getting the deploy to work. 

I had a couple of specific issues with NPM around preinstall hooks firing at the correct time, and the SHA512 changing on the imported the tgz. Both of these weren't issues using yarn, but moving to yarn is a much bigger decision so I worked around them instead for now (they would probably be addressed in future versions of NPM anyway)

Still in draft because it could use some polishing, but ready for people's thoughts

TODO:

Would probably move some of the pack / import logic into build tasks for the 2 packages, rather than install hooks / github workflow tasks like in this POC


GPaoloni (commented):
Didn't knew about this, really great stuff :rocket:. Thanks!! 
I'm 100% in with going this way, it's cleaner and simpler.

PR #334, opened 2022-10-17
update workflow branch

acedeywin:
## Description

- Update Workflow

### Checklist
- [ ] New tests added

### Related Issues
Fixes CHI-1259


.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -14,121 +14,121 @@ on:
...
+    uses: techmatters/serverless/.github/workflows/uk_staging.yml@master
     secrets: inherit
     with:
       changelog: ${{ inputs.changelog }}
      
       
 
   deploy-ukraine_staging: 
-    uses: techmatters/serverless/.github/workflows/ukraine_staging.yml@CHI-1259_workflow_to_deploy_serverless_helpline_stg_norerun
+    uses: techmatters/serverless/.github/workflows/ukraine_staging.yml@master
```
stephenhand, line 117: Can you update this to `hungary_staging.yml` as well?

stephenhand (commented):
Do we have to set the branch explicitly branch to @master? Can we not leave it as the default?

If we deploy from a tag (which will be typical going forward) and the workflow on master has changed, this could unexpected behaviour

@GPaoloni - do we have to specify a branch here if the workflow is from the same repo as the parent?

.github/workflows/deploy-helplines-parallel-worflows.yml
acedeywin: @stephenhand It's done.

stephenhand (approved):
LGTM

PR #343, opened 2022-10-25
CHI-1458: Fix QA release version bumps

stephenhand:
@GPaoloni 

## Description

Change serverless workflow to pass in `github.repository` when generating a release - will allow QA build revision bumps to work correctly

### Checklist
- N / A

### Related Issues
Fixes CHI-1458
### Verification steps

1. Release a version which already has a QA release
2. Check the released version is 1 QA revision higher than the previous highest revision


GPaoloni (approved):
Not sure what's the difference between those two but I know you tested so LGTM!

PR #349, opened 2022-11-01
chore: hardcode port for local dev

robert-bo-davis:
Primary Reviewer: @stephenhand 

## Description
This hardcodes the port for serverless in local dev so that we can simplify spinning up local dev environments and updates nvmrc. Sibling PR coming in techmatters/flex-plugins#1009 to use the new port for local dev.

### Verification steps
run "npm start:local" locally and verify that serverless starts on port 3030


README.md
```diff
@@ -1,64 +1,70 @@
...
-4- run typescript compiler (as Twilio ST serves the .js files) and start local server:  
-`npm start`  
-
-For help on twilio-run commands run:  
-`npm run tr -- help`  
-   
+1- Clone repository:
+`git clone https://github.com/tech-matters/serverless && cd serverless`
+2- Install dependencies:
+`npm ci`
```
stephenhand, line 21: Unfortunately, we've recently learned the Twilio Serverless deployment infrastructure doesn't respect package-lock.json, so if anything, it's probably safer to use `npm install` locally - at least we will be using something likely to be close to the same set of versions that end up being deployed.

Since I found out this, I'm questioning whether we should check in the package-lock.json at all for this repo

I suppose we could try doing something fancy to work around this, like generating a 'deployment only' package.json with a fully expanded set of dependencies, all with explicit versions - based on the package-lock.json file, whilst keeping our package.json as it is. But this sounds non-trivial. In the meantime, really not sure whether checking in a package-lock.json for this repo is helpful.

stephenhand (approved):
LGTM. Some random musings in a comment but nothing blocking

README.md
robert-bo-davis: @stephenhand 

I'd say the package.lock may still have value (along with using `npm ci` locally) to make sure that versions in devDependencies are standardized across local dev envs and so that dependabot will work.

Should we put in a jira ticket for locking versions in `dependencies` by removing `^` and such at a later date? Or should I pull the versions that are currently in the lock file and do that quickly as part of this?

stephenhand: Good point about dependabot - although if it's running against different versions of deps to the ones we actually deploy... Still, best efforts etc.

Removing the version range markers will help, but I think transitive dependencies will still not be guaranteed to be the same version so it's not a complete solution.

Better than nothing though - if you don't mind pinning the versions as part of this PR that would be great

stephenhand (approved):

PR #352, opened 2022-11-03
adding custom actions for Te Guio Production: Twitter and Instagram

janorivera:
Primary reviewer
@GPaoloni  or @stephenhand 

## Description

Adding Twitter and Instagram channels to Te Guío CO production release.

### Checklist
- [na] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/.github/actions/custom-actions/action.yml
+++ b/.github/actions/custom-actions/action.yml
@@ -10,47 +10,47 @@ runs:
   using: "composite"
   steps:
     - name: Childline Thailand Staging release custom action
-      if: github.workflow == format('Thailand Staging release')
+      if: inputs.helpline-name == format('TH_STG')
       uses: ./.github/actions/custom-actions/thailand_staging_custom
     - name: Aarambh Staging release custom action
       if: inputs.helpline-name == format('IN_STG')
       uses: ./.github/actions/custom-actions/aarambh_staging_custom
     - name: Aarambh Production release custom action
-      if: github.workflow == format('Aarambh Production release')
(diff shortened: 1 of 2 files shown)


.github/actions/custom-actions/action.yml
```diff
@@ -45,6 +45,9 @@ runs:
     - name: TeGuio Colombia Staging release custom action
       if: inputs.helpline-name == format('CO_STG')
       uses: ./.github/actions/custom-actions/teguio_staging_custom
+    - name: TeGuio Colombia Production release custom action
+      if: inputs.helpline-name == format('Te Guío CO Production release')
```
stephenhand, line 49:
```suggestion
      if: inputs.helpline-name == format('CO_PROD')
```

You could also do 

```suggestion
      if: github.workflow == format('Te Guío CO Production release')
```

But the first one is more in line with what we want to do in the future. In fact if you wanted to migrate the other production custom action checks for production helplines over to checking ìnput.helpline-name` that would save somebody a job in the future, but np if you prefer not to

stephenhand (commented):
One small issue, otherwise looks ok

.github/actions/custom-actions/action.yml
janorivera: @stephenhand  alright! is this the only change needed? I updated the other values per your comment, but I'm unsure if I need to update anything else.  (I don't fully understand these workflows, I don't know for example from where the inputs.helpline-name is getting its value :) )

stephenhand (approved):
Thanks @janorivera

PR #354, opened 2022-11-07
Localizing postSurvey "good bye" message

GPaoloni:

## Description
This PR adds the ability to localize the "post survey complete". The way this is done is, the original (contact) task contains a `language` attribute, it will be passed down to the post survey task when is created (`/postSurveyInit`). This `language` property is retrieved when the survey is completed (`/postSurveyComplete`) and used to retrieve the translation, if it exists.

### Checklist
- [x] [Corresponding issue has been opened](https://bugs.benetech.org/browse/CHI-1371)
- [ ] New tests added

### Verification steps
Note: this is deployed to development to make testing easier,
- Start a new chat contact and accept it from Flex.
- Edit the generated task from Twilio Console, adding the attribute `"language": "es"`.
- Complete the task from Flex.
- Take the post survey from the contact's view and confirm that the last message (when the post survey is done) is translated.

### DIFF
--- a/functions/postSurveyComplete.protected.ts
+++ b/functions/postSurveyComplete.protected.ts
@@ -148,9 +148,9 @@ export const handler: ServerlessFunctionSignature<EnvVars, Event> = async (
           ) as OneToManyConfigSpec[];
 
           const surveyTaskAttributes = JSON.parse(surveyTask.attributes);
-
+Object.entries(([k, v]) => console.log(k, JSON.stringify(v)))
           // Set the taskLanguage carried over from the original task, if any
-          taskLanguage = surveyTaskAttributes.taskLanguage;
+          taskLanguage = surveyTaskAttributes.language;
 
           // parallel execution to save survey collected data in insights and hrm
           await Promise.all([
--- a/functions/postSurveyComplete.protected.ts
+++ b/functions/postSurveyComplete.protected.ts
@@ -56,7 +56,6 @@ const saveSurveyInInsights = async (
     surveyTaskAttributes,
     memory,
   );
-  console.log('finalAttributes: ', JSON.stringify(finalAttributes));
 
   await surveyTask.update({ attributes: JSON.stringify(finalAttributes) });
 };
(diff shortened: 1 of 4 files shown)


functions/postSurveyComplete.protected.ts
```diff
@@ -88,20 +91,37 @@ const saveSurveyInHRM = async (
   });
 };
 
+const getPostSurvetCompleteMessage = (taskLanguage: string | undefined): string => {
```
stephenhand, line 94: Typo - `getPostSurveyCompleteMessage`

functions/postSurveyComplete.protected.ts
```diff
@@ -88,20 +91,37 @@ const saveSurveyInHRM = async (
...
+        Runtime.getAssets()[`/translations/${taskLanguage}/postSurveyMessages.json`].open(),
+      );
+
+      if (translation.postSurvetCompleteMessage) return translation.postSurvetCompleteMessage;
+    } catch {
+      console.error(`Couldn't retrieve postSurvetCompleteMessage translation for ${taskLanguage}`);
+    }
+  }
+
+  return 'Thank you for reaching out. Please contact us again if you need more help.';
```
stephenhand, line 107: Should we have the default behaviour use the `en-US` file rather than a hardcoded value if the task language is not set?

If we update the standard message, we might only remember to change it in the translations and forget about the hardcoded fallback

stephenhand (approved):
LGTM, v minor comments

functions/postSurveyComplete.protected.ts
```diff
@@ -45,13 +45,17 @@ const saveSurveyInInsights = async (
...
     .buildSurveyInsightsData as BuildSurveyInsightsData;
 
-  const taskAttributes = JSON.parse(surveyTask.attributes);
-  const finalAttributes = buildSurveyInsightsData(postSurveyConfigJson, taskAttributes, memory);
+  const finalAttributes = buildSurveyInsightsData(
+    postSurveyConfigJson,
+    surveyTaskAttributes,
+    memory,
+  );
   console.log('finalAttributes: ', JSON.stringify(finalAttributes));
```
nickhurlburt, line 59: I know this isn't part of this PR -- but will this `console.log` possibly trigger pages?

GPaoloni: No, `console.log`s do not trigger the reporter, only `warn`s and `error`s. But I will remove this line anyways.

functions/postSurveyComplete.protected.ts
GPaoloni: My thinking here was that `en-US` might actually be language that's set. I tried to always leave the default message as something hard coded, in case for some reason we can't get the assets, at least the user will receive some message. Happy to change if it's preferred to rely on the jsons exclusively tho!

stephenhand: I forgot these translations aren't local to the code. Maybe fall back to en-US and then fall back to a hardcoded string. I'm just aiming for consistency with other translation behaviour

PR #360, opened 2022-11-24
Chi 1405 backend access iwf self generated report api

acedeywin:
## Description

- New feature: CSAM: Build backend to provide access to IWF Self-Generated Report URL API

### Checklist
- [x] New tests added

- [x] Strings are localized

### Related Issues
Fixes CHI-1405

### DIFF
--- a/.github/actions/custom-actions/aselo_development_custom/action.yml
+++ b/.github/actions/custom-actions/aselo_development_custom/action.yml
@@ -95,12 +95,15 @@ runs:
     - name: Add IWF_API_URL
       run: echo "IWF_API_URL=${{ env.IWF_API_URL }}" >> .env
       shell: bash
-    # - name: Add IWF_API_ENVIRONMENT
-    #   run: echo "IWF_API_ENVIRONMENT=${{ env.IWF_API_ENVIRONMENT }}" >> .env
-    #   shell: bash
-    # - name: Add IWF_API_COUNTRY_CODE
-    #   run: echo "IWF_API_COUNTRY_CODE=${{ env.IWF_API_COUNTRY_CODE }}" >> .env
-    #   shell: bash
+    - name: Add IWF_SECRET_KEY
+      run: echo "IWF_SECRET_KEY=${{ env.IWF_SECRET_KEY }}" >> .env
+      shell: bash
+    - name: Add IWF_API_CASE_URL
+      run: echo "IWF_API_CASE_URL=${{ env.IWF_API_CASE_URL }}" >> .env
+      shell: bash
+    - name: Add IWF_REPORT_URL
+      run: echo "IWF_REPORT_URL=${{ env.IWF_REPORT_URL }}" >> .env
+      shell: bash
 
     - name: Add TWITTER_CONSUMER_KEY
       run: echo "TWITTER_CONSUMER_KEY=${{ env.TWITTER_CONSUMER_KEY }}" >> .env
--- a/.github/actions/custom-actions/aselo_development_custom/action.yml
+++ b/.github/actions/custom-actions/aselo_development_custom/action.yml
@@ -29,7 +29,7 @@ runs:
       #   with:
       #     ssm_parameter: "DEV_AS_IWF_API_COUNTRY_CODE"
       #     env_variable_name: "IWF_API_COUNTRY_CODE"
-      
+
     - name: Set IWF_SECRET_KEY
       uses: "marvinpinto/action-inject-ssm-secrets@latest"
       with:
@@ -112,6 +112,7 @@ runs:
     - name: Add IWF_API_URL
       run: echo "IWF_API_URL=${{ env.IWF_API_URL }}" >> .env
       shell: bash
+      
     - name: Add IWF_SECRET_KEY
       run: echo "IWF_SECRET_KEY=${{ env.IWF_SECRET_KEY }}" >> .env
       shell: bash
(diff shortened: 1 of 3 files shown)


.github/actions/custom-actions/aselo_development_custom/action.yml
```diff
@@ -29,6 +29,23 @@ runs:
       #   with:
       #     ssm_parameter: "DEV_AS_IWF_API_COUNTRY_CODE"
       #     env_variable_name: "IWF_API_COUNTRY_CODE"
+
+    - name: Set IWF_SECRET_KEY
```
stephenhand, line 33: Will the SSM parameters need to be scoped to the Twilio account, at least the secret key, looks like the report URL too? e.g., `DEV_IWF_AS_SECRET_KEY`

functions/selfReportToIWF.ts
```diff
@@ -0,0 +1,63 @@
...
+    const response = responseWithCors();
+    const resolve = bindResolve(callback)(response);
+
+    try {
+      const { user_age_range } = event;
+      if (!user_age_range) return resolve(error400('user_age_range'));
+
+      const body: IWFSelfReportPayload = {
+        secret_key: context.IWF_SECRET_KEY,
+        case_number: v4(),
```
stephenhand, line 41: Do we need to record this anywhere on our end? @GPaoloni

tests/selfReportToIWF.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  });
+  afterAll(() => {
+    helpers.teardown();
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('Should return status 400', async () => {
```
stephenhand, line 50: Can you name the test to specify the circumstances under which it should return a 400? Same for all the other tests

functions/selfReportToIWF.ts
```diff
@@ -0,0 +1,63 @@
...
+        },
+        validateStatus: () => true,
+      });
+
+      if (report.data?.result !== 'OK') return resolve(error400(report.data?.message));
+
+      const reportUrl = `${context.IWF_REPORT_URL}/t?=${report.data?.message?.access_token}`;
+      return resolve(send(report.data?.result)(reportUrl));
+    } catch (error) {
+      return resolve(error500(error as any));
```
stephenhand, line 60: Does the error response returned here make it easy to distinguish which HTTP call failed, the first or the second? We might need to know if we successfully called the API but failed to report it at some point

stephenhand (commented):
Few comments, the only major one being re: SSM parameter naming

.github/actions/custom-actions/aselo_development_custom/action.yml
robert-bo-davis: Is this a new key that will be created? Does it make sense to go ahead and start using the new SSM naming convention to reduce the number of things that need to migrate?

I'm happy to make some suggestions about the structure if so. I'll need to ask some questions about the context of the secret to make those, though.

functions/selfReportToIWF.ts
acedeywin: @stephenhand, please, can you throw more light on this?

stephenhand: My misunderstanding please disregard

tests/selfReportToIWF.test.ts
acedeywin: It's done.

.github/actions/custom-actions/aselo_development_custom/action.yml
stephenhand: @robert-bo-davis - sorry I missed this - we'll use the legacy convention for this case but start using the new one going forward

robert-bo-davis: @stephenhand No worries. Stephen and I had already discussed and come to the same conclusion.

.github/actions/custom-actions/aselo_development_custom/action.yml
```diff
@@ -95,12 +112,16 @@ runs:
...
       shell: bash
-    # - name: Add IWF_API_ENVIRONMENT
-    #   run: echo "IWF_API_ENVIRONMENT=${{ env.IWF_API_ENVIRONMENT }}" >> .env
-    #   shell: bash
-    # - name: Add IWF_API_COUNTRY_CODE
-    #   run: echo "IWF_API_COUNTRY_CODE=${{ env.IWF_API_COUNTRY_CODE }}" >> .env
-    #   shell: bash
+
+    - name: Add AS_DEV_IWF_SECRET_KEY
+      run: echo "AS_DEV_IWF_SECRET_KEY=${{ env.AS_DEV_IWF_SECRET_KEY }}" >> .env
```
GPaoloni, line 117: This will deploy an env var called `AS_DEV_IWF_SECRET_KEY` to the actual Serverless instance. Is that intended? How will the code know which variable to use on each account?

functions/selfReportToIWF.ts
GPaoloni: Yes, I think we want to use the one generated in HRM for this. @acedeywin you can check that Flex, before calling this function [here](https://github.com/techmatters/flex-plugins/blob/0e2c8109a53462d6fcb912a4f4f5c4a263904337/plugin-hrm-form/src/components/CSAMReport/CSAMReport.tsx#L137), already have access to a csam record, which [contains a UUID generated under `.csamReportId` property](https://github.com/techmatters/hrm/blob/810e8732f6858058f6529d1f04ddbe1b99666ac0/hrm-service/src/csam-report/csam-report.ts#L22).

GPaoloni (commented):
Small question, and replying to the thread where I was mentioned!

.github/actions/custom-actions/aselo_development_custom/action.yml
acedeywin: @GPaoloni, the env is for staging, which I believe is what the Aselo Development was created for. The prod env will be deployed to the prod account of helplines that would use the feature in production. @stephenhand, I am not sure if I answered correctly.

PR #361, opened 2022-11-25
Mt stg workflow

Humairaa127:
## Description
Add Malta Kellimni deployment workflow file
Add workflow file to deploy-parallel-helplines workflow


.github/workflows/deploy-helplines-parallel-worflows.yml
```diff
@@ -120,6 +120,12 @@ jobs:
     secrets: inherit
     with:
       changelog: ${{ inputs.changelog }}
+      
+  deploy-kellimni_staging:
+    uses: techmatters/serverless/.github/workflows/Kellimni_staging.yml
```
GPaoloni, line 125: Usually we lowercase the file names.

GPaoloni (approved):
Looks good! 
Since it's a new helpline, I assume it does not needs anything from the "custom actions" for now, right?

PR #362, opened 2022-11-28
remapping post survey answers to match datatypes

janorivera:
Reviewer
@Humairaa127 
## Description

I'm updating this because I think that the type of answers don't match the current data type for the Twilio attribute.

### DIFF
--- a/assets/formDefinitions/co-v1/insights/postSurvey.private.json
+++ b/assets/formDefinitions/co-v1/insights/postSurvey.private.json
@@ -1,7 +1,7 @@
 [
   {
     "insightsObject": "conversations",
-    "attributeName": "conversation_attribute_9",
+    "attributeName": "conversation_measure_4",
     "questions": ["didUserFeelSupported"]
   },
   {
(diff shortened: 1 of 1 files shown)


Humairaa127:
The flex insights config is already using this attribute, so that needs to be adjusted if there are any left:
     {
        "name": "memberOfAnEthnic",
        "insights": ["conversations", "conversation_measure_4"],
        "type": "mixed-checkbox"
      }

Humairaa127 (changes requested):
The flex insights config is already using this one attribute, so that needs to be adjusted if there are any left:
{
"name": "memberOfAnEthnic",
"insights": ["conversations", "conversation_measure_4"],
"type": "mixed-checkbox"
}

janorivera:
@Humairaa127 yes! I'm preparing a PR in flex-plugins with all the proper adjustments. I have a call with them today to have a final discussion about the things that they want to store and I'll raise the PR so you can review it.

Humairaa127 (approved):

PR #372, opened 2022-12-12
CHI-1578: Queue Transfer

stephenhand:
## Description

* Updates behaviour of chat transfers to treat transfers to queues differently to transfers to counsellors
* Move .prettierrc settings into package.jsons because a bug in webstorm means it doesn't pick them up from .prettierrc

The updated logic means that a 'queue' target transfer will complete the original task as soon as it hits a queue, rather than holding it open until accepted by a worker, which worker transfers do

It also limits the special logic for rejections to worker transfers. Queue transfers just go back on the queue if rejected now

### Checklist
- [x] New tests added

### Related Issues
Fixes CHI-1578

### Verification steps

See ticket

### DIFF
--- a/functions/taskrouterListeners/transfersListener.private.ts
+++ b/functions/taskrouterListeners/transfersListener.private.ts
@@ -45,7 +45,7 @@ const isChatTransfer = (
   taskAttributes.transferMeta.mode === 'COLD' &&
   taskAttributes.transferMeta.transferStatus === 'accepted';
 
-const isChatTransferAccepted = (
+const isChatTransferToWorkerAccepted = (
   eventType: EventType,
   taskChannelUniqueName: string,
   taskAttributes: ChatTransferTaskAttributes,
@@ -54,7 +54,7 @@ const isChatTransferAccepted = (
   isChatTransfer(taskChannelUniqueName, taskAttributes) &&
   taskAttributes.transferTargetType === 'worker';
 
-const isChatTransferRejected = (
+const isChatTransferToWorkerRejected = (
   eventType: EventType,
   taskChannelUniqueName: string,
   taskAttributes: ChatTransferTaskAttributes,
@@ -66,7 +66,7 @@ const isChatTransferRejected = (
 const isChatTransferToQueueComplete = (
   eventType: EventType,
   taskChannelUniqueName: string,
-  taskAttributes: { transferMeta?: TransferMeta; transferTargetType: 'worker' | 'queue' },
+  taskAttributes: ChatTransferTaskAttributes,
 ) =>
   eventType === TASK_QUEUE_ENTERED &&
   isChatTransfer(taskChannelUniqueName, taskAttributes) &&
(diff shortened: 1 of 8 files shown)


functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -29,6 +31,11 @@ type TransferMeta = {
   transferStatus: 'transferring' | 'accepted' | 'rejected';
 };
 
+type ChatTransferTaskAttributes = {
+  transferMeta?: TransferMeta;
+  transferTargetType: 'worker' | 'queue';
+};
+
 const isChatTransfer = (
```
GPaoloni, line 39: Should `taskAttributes.transferMeta.transferStatus === 'accepted'` be removed from this function? Or renamed the function?

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -41,16 +48,29 @@ const isChatTransfer = (
...
 
 const isChatTransferRejected = (
   eventType: EventType,
   taskChannelUniqueName: string,
-  taskAttributes: { transferMeta?: TransferMeta },
+  taskAttributes: ChatTransferTaskAttributes,
 ) =>
   (eventType === RESERVATION_REJECTED || eventType === RESERVATION_TIMEOUT) &&
-  isChatTransfer(taskChannelUniqueName, taskAttributes);
+  isChatTransfer(taskChannelUniqueName, taskAttributes) &&
```
GPaoloni, line 63: Won't this assume that `taskAttributes.transferMeta.transferStatus === 'accepted'`?

GPaoloni (approved):
Changes lgtm. Just small questions about when we consider the transfer "accepted" or "rejected".

functions/taskrouterListeners/transfersListener.private.ts
stephenhand: So apparently in flex, we start all our cold transfers with a status of 'accepted', which is causing me some headaches for determining whether to show the transfer buttons or not... 

We could start using the 'transferring' state for cold transfers, but then we would have to remove the check here. Could that have unexpected ramifications?

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -41,16 +48,29 @@ const isChatTransfer = (
...
+  taskAttributes.transferTargetType === 'worker';
+
+const isChatTransferToQueueComplete = (
+  eventType: EventType,
+  taskChannelUniqueName: string,
+  taskAttributes: { transferMeta?: TransferMeta; transferTargetType: 'worker' | 'queue' },
```
murilovmachado, line 69: Here you can use the type you created: `ChatTransferTaskAttributes`

functions/taskrouterListeners/transfersListener.private.ts
```diff
@@ -41,16 +48,29 @@ const isChatTransfer = (
 const isChatTransferAccepted = (
   eventType: EventType,
   taskChannelUniqueName: string,
-  taskAttributes: { transferMeta?: TransferMeta },
-) => eventType === RESERVATION_ACCEPTED && isChatTransfer(taskChannelUniqueName, taskAttributes);
+  taskAttributes: ChatTransferTaskAttributes,
+) =>
+  eventType === RESERVATION_ACCEPTED &&
+  isChatTransfer(taskChannelUniqueName, taskAttributes) &&
+  taskAttributes.transferTargetType === 'worker';
```
murilovmachado, line 55: Maybe renaming these functions to `isChatTransferToWorkerAccepted` and `isChatTransferToWorkerRejected`?

murilovmachado (approved):
Great to see the changes were "simple". I was worried it could be messier, but I was wrong. Also, checked the behavior and it seems fine.

PR #381, opened 2022-12-20
Add ability to save Contact ID to Insights

GPaoloni:
Primary reviewer: @stephenhand

This PR is related to https://github.com/techmatters/flex-plugins/pull/1138.

WIP:
- [x] CI
- [ ] Add tests

## Description
This PR splits the `assignOfflineContact` function into two:
- `assignOfflineContact` (could be renamed to `assignOfflineContactInit`): creates the task that will be used to link to the contact. Also adds a two minutes timeout to the task to prevent the user having stuck tasks.
- `assignOfflineContactResolve`: receives two possible actions to execute:
  - `complete`: given a `taskSid` and a `finalTaskAttributes`, updates the task attributes and then completes it. This is so the attributes make it to insights.
  - `remove`: removes the task with sid `taskSid` to cleanup if something went wrong. 

This PR makes sense in the context of the frontend changes to achieve "add ability to save Contact ID to Insights".

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1498)
- [WIP] New tests added

### Verification steps
See https://github.com/techmatters/flex-plugins/pull/1138.


stephenhand (commented):
Repeated from our slack convo - LGTM but needs to be backwards compatbile.

You could just leave it as one function, look for both old & new attribute names, and decide whether or not to complete the task based on which attribute was used

stephenhand (approved):
LGTM

PR #384, opened 2023-01-03
Extend operating hours function to support multiple office config

GPaoloni:
## Description
This PR adds support for handling multiple offices operating hours within a single account. To accomplish this, the structure of the expected  `assets/operatingInfo/xxxxxx.private.json` file is slightly different. Reading the code changes in this PR should help (bundled changes for `aselo-dev` as an example).

Previously the operating shifts were specified per-channel, with a structure like
```
type OperatingInfo = {
  timezone: string;
  holidays: { [date: string]: string }; 
  operatingHours: { [channel: string]: { [day in DaysOfTheWeek]: OperatingShift[] } };
};
```
Here we rename this structure to `OperatingInfoEntry`.

The root of the file will have a shape like

```
type OperatingInfo = OperatingInfoEntry & {
  offices: {
    [office: string]: OperatingInfoEntry;
  };
};
```
This means that the previous shape is still valid, and is in fact the "default behavior". On top of that, an extra `offices` object is accepted, where each office can specify it's custom shifts per-channel, holidays and timezone.
If for a given office, there's no entry for a particular channel, it will default to the root specification.
If there's no value for `office`, the root level will be used.

On top of the code changes, to make this work we need to specify the `office` when calling operating info in Studio Flows. So far, the way each office gets routed is based on the `twilioNumber` property of each message (i.e. to which number this message is being sent).
For this purpose, I've added an example on how to do this in Aselo Development -> Messaging Flow. 

Screenshot from 2023-01-03 15-03-01

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1448)
- [x] New tests added

### Verification steps
- Make sure the code of this PR is deployed to Aselo Dev Serverless.
- Make sure that Messaging Studio Flow contains the configuration as shown here (at the time of writing, revision [893](https://www.twilio.com/console/studio/flows/FWf4b5ac8e9bb0ad564033ab4dd4bb5f48/revisions/893) is good).
- Send a FB message while the `office` for it is being set as `Fake Office` and confirm that the message is routed properly (is open).
- Change the "set office" widget in the Studio Flow to make it something other than `Fake Office`, and confirm that you can't start a new contact (since it's defaulting to the root level of the operating info file and `facebook` channel is intentionally always closed)


murilovmachado (approved):
Looks good to me, Gian! (sorry for the delay on this one, I've missed it)

functions/operatingHours.ts
```diff
@@ -21,25 +21,85 @@ enum DaysOfTheWeek {
   Sunday = 7,
 }
 
-type OperatingInfo = {
+type OperatingInfoEntry = {
```
stephenhand, line 24: `OfficeOperatingInfo` might be more descriptive?

functions/operatingHours.ts
```diff
@@ -21,25 +21,85 @@ enum DaysOfTheWeek {
...
+    return 'open';
+  }
+
+  return 'closed';
+};
+
+const getOperatingStatus = (operatingInfo: OperatingInfo, channel: string, office?: string) => {
+  const { offices, ...operatingInfoRoot } = operatingInfo;
+
+  // TODO: should we loudly error if office is provided but there's no operatingInfo.offices entry?
```
stephenhand, line 85: I would say falling back to the default is fine in either of these cases, it will probably make updates easier to coordinate if things don't blow up when there's a transient mismatch (e.g. you've updated the asset but not the flow yet or vice versa)

stephenhand (approved):
Nothing blocking, LGTM

PR #390, opened 2023-01-18
CHI-1484: Self signed certs flag

stephenhand:
## Description

* Self signed certs are only allowed when a flag is set
* Renamed SSM parameters to match new convention
* Aselo Development flagged to allow self signed certs
* Rati Prod & Staging configured for self reported CSAM (staging allows self signed HTTPS)

TODO: Update SSM parameters

### Checklist
- [X] New tests added

### Related Issues
CHI-1484

### Verification steps

* Deploy to Rati Staging & Aselo Development & regression test self reported CSAM

Deploy to prod cannot be tested

### DIFF
--- a/.github/actions/custom-actions/aarambh_production_custom/action.yml
+++ b/.github/actions/custom-actions/aarambh_production_custom/action.yml
@@ -34,6 +34,21 @@ runs:
       with:
         ssm_parameter: "PROD_IN_IWF_API_CHANNEL_ID"
         env_variable_name: "IWF_API_CHANNEL_ID"
+    - name: Set IWF_SECRET_KEY
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: /production/serverless/${{inputs.account-sid}}/iwf_secret_key
+        env_variable_name: "IWF_SECRET_KEY"
+    - name: Set IWF_API_CASE_URL
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: /production/serverless/${{inputs.account-sid}}/iwf_api_case_url
+        env_variable_name: "IWF_API_CASE_URL"
+    - name: Set IWF_REPORT_URL
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: /production/serverless/${{inputs.account-sid}}/iwf_report_url
+        env_variable_name: "IWF_REPORT_URL"
     # Append environment variables
     - name: Add IWF_API_USERNAME
       run: echo "IWF_API_USERNAME=${{ env.IWF_API_USERNAME }}" >> .env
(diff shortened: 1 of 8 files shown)


.github/actions/custom-actions/aselo_development_custom/action.yml
```diff
@@ -1,18 +1,22 @@
...
+    required: true
 runs:
   using: "composite"
   steps:
     # Set IWF credentials
     - name: Set IWF_API_USERNAME
       uses: "marvinpinto/action-inject-ssm-secrets@latest"
       with:
-        ssm_parameter: "IWF_API_USERNAME"
+        ssm_parameter: IWF_API_PASSWORD
```
GPaoloni, line 14: I this change intended? Looks like it's setting local env var `IWF_API_USERNAME` with the value of `IWF_API_PASSWORD`.

GPaoloni (approved):
Looks good to me!

.github/actions/custom-actions/aarambh_production_custom/action.yml
```diff
@@ -34,6 +34,21 @@ runs:
       with:
         ssm_parameter: "PROD_IN_IWF_API_CHANNEL_ID"
         env_variable_name: "IWF_API_CHANNEL_ID"
+    - name: Set IWF_SECRET_KEY
+      uses: "marvinpinto/action-inject-ssm-secrets@latest"
+      with:
+        ssm_parameter: /production/serverless/${{inputs.account-sid}}/iwf_secret_key
```
robert-bo-davis, line 40: To clarify the standard, it is : `/${environment}/${serviceSecretBelongsToNotServiceThatIsUsingIt}/${twilioAccountSid}/${paramName}`

The first two paths are both a permission and a dynamic loading barriers. So we use them to say "this service has access to all of the the secrets for foo service under bar environment. Or "I want to load all of the secrets for foo service in bar environment using `getParametersByPath`. Using the service that is requesting the param instead of the service that the param relates to could also solve this, but may end up requiring duplication of params.

So in this case it would probably be `/production/iwf/${{inputs.account-sid}}/secret_key` if there are going to be different keys per twilio account. If it is global secret that changes based on env it would probably only be `/production/iwf/secret_key` And we haven't really defined a format for global secrets, but I would propose something like `/global/iwf/secret_key`.

robert-bo-davis (changes requested):

.github/actions/custom-actions/aarambh_production_custom/action.yml
robert-bo-davis: Since this is a bit of a priority fix, I'm happy to approve as is and refactor in a next iteration as well so long as we don't let the fix linger too long.

stephenhand: Updated

robert-bo-davis (approved):

PR #392, opened 2023-01-19
CHI-1484: Fix form data stream state in selfReportIwf

stephenhand:
## Description

An issue was occurring where requests were 'hanging' after the first one sometimes for selfReportIWF
This was because the 'FormData' instance used by the function was defined externally. 'FormData' is a stream buffer intended for a single use in a single request. 

It being defined externally resulted in situations where requests were trying to reuse a closed stream and hanging

* Move form data variable inside function scope
* Fix tests

### Checklist
- [X] New tests added

### Related Issues
Fixes CHI-1484

### Verification steps

Deploy the fix & regression test self reported CSAM

### DIFF
--- a/functions/selfReportToIWF.ts
+++ b/functions/selfReportToIWF.ts
@@ -31,8 +31,6 @@ export type Body = {
   request: { cookies: {}; headers: {} };
 };
 
-export const formData = new FormData();
-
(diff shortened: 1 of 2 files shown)


acedeywin (approved):
LGTM

PR #394, opened 2023-01-24
CHI-1657: Disable operating hours checks for non prod builds by default

stephenhand:
Note: this was kind of a 'rage PR' that I raised after the staging helpline I was trying to debug refused to let me initiate a webchat because it was 'closed for the day'

So it may not be implemented in the most optimal manner, appreciate your input!

## Description

It is extremely frustrating trying to debug an account specific issue in staging only to be blocked out of chat by an operating hours restriction. Operating hours should not be enforced by default for non production builds, only when you want to test that particular functionality

* Disable operating hours checks for non prod builds by default
* With an override option

### Checklist
- N/A New tests added

### Related Issues
CHI-1657

### Verification steps

* Deploy a staging build, check that the channels can be used out of hours
* Deploy a staging build with the override checked, ensure operating hours are enforced
* Deploy a prod build without the override checked, ensure operating hours are enforced

### DIFF
--- a/.github/workflows/custom_helpline.yml
+++ b/.github/workflows/custom_helpline.yml
@@ -15,6 +15,8 @@ on:
       environment_code:
         description: The short upper case code used to identify the environment internally, e.g. STG, PROD, DEV
         required: true
+      force_enable_operating_hours:
+        description: Forces operating hours to be enforced for this deployment (by default they are only enforced in PROD)
   workflow_call:
     inputs:
       helpline_code:
@@ -146,3 +148,4 @@ jobs:
           aws-region: $AWS_REGION
           s3-bucket: $S3_BUCKET
           helpline-name: ${{inputs.helpline_code}}_${{inputs.environment_code}}
+          disable-operating-hours: ${{ (inputs.force_enable_operating_hours || inputs.environment_code == 'PROD') && 'false' || 'true' }}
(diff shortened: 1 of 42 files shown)


.github/workflows/custom_helpline.yml
```diff
@@ -146,3 +148,4 @@ jobs:
           aws-region: $AWS_REGION
           s3-bucket: $S3_BUCKET
           helpline-name: ${{inputs.helpline_code}}_${{inputs.environment_code}}
+          disable-operating-hours: ${{ (inputs.force_enable_operating_hours || inputs.environment_code == 'PROD') && 'false' || 'true' }}
```
GPaoloni, line 151: This line got me thinking a while. Maybe a comment that explains a bit is useful here?

GPaoloni (approved):
LGTM. The `DISABLE_OPERATING_HOURS_CHECK` env var can be flipped from Twilio Console if we really need to test operating hours in staging, as another option that does not even requires PRs. I suggest tagging the solutions team so they are aware of this.

stephenhand:
> LGTM. The `DISABLE_OPERATING_HOURS_CHECK` env var can be flipped from Twilio Console if we really need to test operating hours in staging, as another option that does not even requires PRs. I suggest tagging the solutions team so they are aware of this.

That's great to hear, I thought these vars could only be updated with deploys

PR #398, opened 2023-01-27
CHI-1646: Incoming Instagram Story Replies are ignored

stephenhand:
**Description**

Any message sent to the InstagramToFlex function identified as being a reply to a story will no longer be forwarded on to Flex

[X] Tests Added
[X] Issue Created

Issue:

CHI-1646

Validation Steps:

Not sure, never used Instagram before, I hope @dee-lou will know

### DIFF
--- a/functions/webhooks/instagram/InstagramToFlex.ts
+++ b/functions/webhooks/instagram/InstagramToFlex.ts
@@ -36,6 +36,12 @@ type InstagramMessageObject = {
   };
 };
 
+type InstagramStoryReply = InstagramMessageObject['message'] & {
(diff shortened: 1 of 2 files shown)


murilovmachado (approved):
Code looks good to me. If anyone could test it, it would be good, though. We'd need to create an Instagram story and reply to it. Is that right? Does anyone have access to the Instagram Account used by `Aselo Dev`?

stephenhand:
Ok, I'll merge it and we can test it as part of QA, I'll chase up with @dee-luo for setting up a test

PR #406, opened 2023-02-01
Fix IWF Self-Gen URL generator format

GPaoloni:
## Description
Previously, for the self-gen reporting URL flow we were generating: `https://<url>/t?=TOKEN  `
Now it will be: `https://<url>/?t=TOKEN`

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1660)
- [x] New tests added

### DIFF
--- a/tests/selfReportToIWF.test.ts
+++ b/tests/selfReportToIWF.test.ts
@@ -103,7 +103,28 @@ describe('selfReportToIWF', () => {
       user_age_range: '<13',
     };
 
-    await selfReportToIWF({ ...baseContext }, event, () => {});
+    // @ts-ignore
+    axios.mockImplementationOnce(async () => ({
+      data: {
+        result: 'OK',
+        message: {
+          access_token: 'SECRET TOKEN',
+        },
+      },
+    }));
+
+    const callback: ServerlessCallback = (err, result) => {
+      expect(result).toBeDefined();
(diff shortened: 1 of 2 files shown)


tests/selfReportToIWF.test.ts
```diff
@@ -103,7 +103,28 @@ describe('selfReportToIWF', () => {
...
+    axios.mockImplementationOnce(async () => ({
+      data: {
+        result: 'OK',
+        message: {
+          access_token: 'SECRET TOKEN',
+        },
+      },
+    }));
+
+    const callback: ServerlessCallback = (err, result) => {
```
stephenhand, line 116: As I said in the other PR, this approach is slightly flawed because it passes if the callback never gets called.

If we make the callback a `jest.fn()`, we can verify they gets called at all using `expect(...).toHaveBeenCalled()`, then run detailed assertions on the parameters by pulling them from the `mocks` property: https://jestjs.io/docs/mock-function-api#mockfnmockcalls (because `expect(...).toHaveBeenCalledWith(...)` can be limiting.

Generally, any assertion checks that gets executed conditionally should be considered a smell in testing code

stephenhand (approved):
LGTM - thanks for beefing up the tests, just one note there

PR #408, opened 2023-02-02
chore: add ignores to license add script

robert-bo-davis:
Adds ignore patterns (not all but the important ones... hopefully?) to the license add command and windows versions.


stephenhand (approved):
Thx!

PR #409, opened 2023-02-02
Support status-based messages to be included in operatingHours response

GPaoloni:
## Description
This PR modifies the following functions:
- `functions/operatingHours.ts`: now supports two new optional parameters in the request: `includeMessageTextInResponse` and `language`. If `includeMessageTextInResponse` is not provided, the function will continue to work as before (prevent breaking changes). If the function is called with `includeMessageTextInResponse` = `true`, the changes on this PR will apply: the response won't be a plain `string` representing the status (`open`, `closed`, `holiday`), but an object of the shape `{ status: string; message: string }`. The message will indicate the reason (if closed due to shifts or holidays). If `language` is provided, we'll attempt to retrieve the translated version, defaulting to `en-US` if not.
- `functions/sendSystemMessage.ts`: now supports providing `taskSid` or `channelSid` as the destination for the message. Exactly one must be provided.
- Adds `functions/sendMessageAndRunJanitor.protected.ts` which is just a wrapper around `sendSystemMessage`, that does not asks for Token validation, but being marked as `.protected` will only accept requests with a valid [X-Twilio-Signature header](https://www.twilio.com/docs/usage/security). This is required so we can reuse the `sendSystemMessage` functionality from the context of a Studio Flow, and there's no `Token` to include in the request. This function also:
  - Removes the `'studio'` type webhooks from the channel to avoid triggering a second execution. This is needed to fix the [bottom section in this description](https://github.com/techmatters/serverless/pull/409#note:~:text=I%27m%20seeing%20an,a%20bit%20hacky.).
  - Invokes the chat channel janitor to perform a cleanup on the channel (not sure if required but won't hurt deactivating the maybe-stale channel).

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1635)
- [x] New tests added

### Verification steps
- Make sure this branch is deployed to Aselo Development.
- Go to Twilio Console -> Functions -> Serverless and manually remove (or set to empty) the environment variable `DISABLE_OPERATING_HOURS_CHECK` if it is present (this way we'll get operating hours checks).
- Make sure that the Messaging Studio Flow is wired like in the picture above.
- Sending messages from Facebook and Instagram should always return "closed due to shift" (because of the shape of `aselo-dev` operating info asset). Messages from webchat should succeed.
- Following the validation steps of https://github.com/techmatters/webchat/pull/184 might also help :).
- Once done testing, either turn set `DISABLE_OPERATING_HOURS_CHECK` back to true or re-deploy.

Studio Flow:

Screenshot from 2023-02-02 17-08-58

The updates in the Studio Flow consist of the following (transitions):
- Determine office (`split office` widget).
- Save the variables of the current office (`set X office`).
- Run function `open hours` with
  - `channel`: `{{flow.variables.channel}}`.
  - `office`: `{{flow.variables.channel}}`.
  - `includeMessageTextInResponse`: true :bangbang: This is new.
  - `language` (optional): the language from where it is stored (pre engagement for webchat, maybe a bot's memory for other channels?
- Determine the response from above using a split widget based on `widgets.openHours_office.parsed.status`. If
  - Contains `open` -> follow normal flow execution.
  - No condition matched means that it is either `closed` or `holiday` -> Run function `/sendMessageAndRunJanitor` with:
    - `channelSid`: `{{flow.channel.address}}`.
    - `message`: `{{widgets.openHours_office.parsed.message}}`.

### Note:
I'm seeing an unexpected behavior, the function to send the message is being called multiple times, resulting in repeated messages being sent to the channels. This is occurring for custom and Twilio channels. I have a support ticket asking for this behavior, but if you have some idea the help is very appreciated.
Update on this: looks like the message being sent by the `system` to the active channel, having a different `from` address is causing a second Studio Flow execution. This is likely the reason why Twilio does not includes the `xTwilioWebhookEnabled = true` for the "send message widget" :) I'm hoping to get some assistance from Support here. A potential idea to prevent this behavior would be to catch every incoming message at the beginning of the Studio Flows, and just "do nothing" if the message is comming from `system`, but feels a bit hacky.


functions/operatingHours.ts
```diff
@@ -128,7 +166,20 @@ export const handler = async (
 
     const status = getOperatingStatus(operatingInfo, channel, office);
 
-    resolve(success(status));
+    // Support legacy function to avoid braking changes
+    // TODO: remove once every account has been migrated
+    if (!useV2) {
```
stephenhand, line 171: I know it's for compatibility and we never expect a caller to explicitly NOT use V2, but it's still a bit weird that passing `'false'` into `useV2` would activate v2... 

Also, sorry to be a pain but can we maybe use a more descriptive name for this flag? Like `includeMessageTextInResponse` or something? 'V2' doesn't really help people understand what the flag is doing.

functions/sendSystemMessage.ts
```diff
@@ -18,54 +17,78 @@ type EnvVars = {
...
-  newStatus?: string;
   request: { cookies: {}; headers: {} };
 };
 
+export const sendSystemMessage = async (context: Context<EnvVars>, event: Body) => {
+  const { taskSid, channelSid, message, from } = event;
+
+  console.log('------ sendSystemMessage excecution ------');
+
+  if (taskSid && channelSid) {
```
stephenhand, line 31: Do we really care if they provide both? I'd say a log stating that both were provided so using `channelSid` in preference is the most we need

functions/sendSystemMessage.ts
```diff
@@ -18,54 +17,78 @@ type EnvVars = {
...
+    return { status: 400, message: 'missing message.' };
+  }
+
+  const client = context.getTwilioClient();
+
+  let channelSidToMessage = null;
+
+  if (channelSid !== undefined) {
+    channelSidToMessage = channelSid;
+  } else if (taskSid !== undefined) {
```
stephenhand, line 52: If check here is redundant - if you loosened up the checks above to be simple truthy / falsey checks I don't think TS would complain either.  None of the falsey values are valid for either of these inputs so it's a better check anyway

functions/sendSystemMessage.ts
```diff
@@ -18,54 +17,78 @@ type EnvVars = {
 
 export type Body = {
```
stephenhand, line 18: If you wanted to be super anal with your types you could do something like (assuming you agree with the comment about allowing both):

```
export type Body = ({
  taskSid?: string;
  channelSid: string} | {
  taskSid: string;
  channelSid?: string}) & {
  message?: string;
  from?: string;
  request: { cookies: {}; headers: {} };
};
```

But what you have is completely fine

stephenhand (commented):
Couple of comments, mostly naming and other nits, only real note is on the tests

stephenhand (approved):
LGTM

PR #426, opened 2023-03-22
Added condition to only close original task if transfer is accepted

GPaoloni:
This PR is related to https://github.com/techmatters/flex-plugins/pull/1307

## Description
While working on the linked PR, after fixing the bug, I've noticed that rejecting a warm voice transfer after joining the conference will also close the original task, causing an ugly bug where the original counselor does not have a way to interact with the call, but the communication was still open with the caller.

This PR adds a check to the warm transfers case in charge of closing the tasks in wrapup state, to only do so if the transfer is `accepted` (either a `COLD` transfer or a fully accepted `WARM` transfer will be in this state).
The rejected `WARM` transfer does enters this case, as the `transferStatus` is `rejected`, and the rejecting side will close the rejected task from the frontend.

### Checklist
- [x] New tests added

### Verification steps
- Make sure that you have a local instance of Flex at the linked PR running.
- Perform a warm transfer of a voice task using as the target the worker at the Flex instance mentioned above.
- Join the conference.
- Reject the transfer.
- Confirm that the original task is not closed and it re-takes control of the forms.

### DIFF
--- a/functions/taskrouterListeners/transfersListener.private.ts
+++ b/functions/taskrouterListeners/transfersListener.private.ts
@@ -99,14 +99,15 @@ const isWarmVoiceTransferRejected = (
   taskAttributes.transferMeta &&
   taskAttributes.transferMeta.mode === 'WARM';
 
-const isVoiceTransferWrapup = (
+const isVoiceTransferOriginalInWrapup = (
   eventType: EventType,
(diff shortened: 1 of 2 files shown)


stephenhand (approved):
LGTM

PR #438, opened 2023-04-13
End post survey task

murilovmachado:
## Description
This PR fixes the bug in which the post-survey task is not ended when calling `endChat`.

The fix:
- Extracted the code that used to end the contact task to a function
- Applied this function to not only the contact task, but to the post-survey task as well
- Handled `channelCleanupRequired` 

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1870

### Verification steps
At webchat, at post-survey stage, the user can hit end chat and the post-survey is ended.

### DIFF
--- a/functions/endChat.ts
+++ b/functions/endChat.ts
@@ -63,6 +63,105 @@ const getEndChatMessage = (event: Body): string => {
   return 'User left the conversation';
 };
 
+/**
+ * End a task by updating their assignment status.
+ *
+ * It also sends a message indicating that the user has left the conversation
+ * if appropriate.
+ *
+ * @returns channelCleanupRequired
+ */
+const updateTaskAssignmentStatus = async (
+  taskSid: string,
+  channelSid: string,
+  context: Context<EnvVars>,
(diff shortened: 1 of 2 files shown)


functions/endChat.ts
```diff
@@ -63,6 +63,105 @@ const getEndChatMessage = (event: Body): string => {
...
+      surveyTaskSid,
+      channelSid as string,
+      context,
+      event,
+    );
+    updateTaskPromises.push(updatePostSurveyTask);
+  }
+
+  const resolvedPromises = await Promise.all(updateTaskPromises);
+  return resolvedPromises.some((channelCleanupRequired: boolean) => channelCleanupRequired);
```
GPaoloni, line 162: I think this can cause race conditions if the original task does not exists any more. I don't think this will make the function to entirely fail, but just in case can we use `Promise.allSettled()` instead?

GPaoloni (approved):
Looks an works great! Just one comment on a possible edge case.

PR #449, opened 2023-05-02
CHI-1972: Fix emojis for line

stephenhand:
## Description

Fixes CHI-1972, emojis in text should now not have signature mismatches. The issue was reported to LINE here: https://github.com/line/line-bot-sdk-php/issues/269 . The issue seems to be that LINE is unnecessarily escaping emoji unicode characters for JSON (they don't need to be escaped) and using that 'escaped' text to generate the signature, whereas `JSON.stringify` that we use in our signature generating algorithm does not escape these characters, so we get different signatures generated. 

I'm not sure how complete this fix is, because I don't know exactly which characters are being unnecessarily escaped. This fix was specifically focused on emojis, but other Unicode characters should be tested.

I also beefed up the unit tests so now the signature matching algorithm isn't stubbed out, and invalid signature cases are tested

TODO: Remove debug logging

### Checklist
- [X] New tests added

### Related Issues
CHI-1972

### Verification steps

* Start a LINE chat on Aselo Development
* Send standard text back & forth, verify it is received
* Send standard emojis back & forth, verify they are received
* Send 'adjusted' emojis back & forth, like emojis with skin tone adjustments etc.
* Send other Unicode characters

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,5 +44,6 @@ module.exports = {
     "implicit-arrow-linebreak": "off",
     "object-curly-newline": "off",
     "function-paren-newline": "off",
+    "linebreak-style": "off",
   },
 };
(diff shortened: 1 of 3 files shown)


stephenhand:
> Looks good to me!
> 
> What an incredible catch. I assumed it was something with the unicode, but never imagined something like this. How did you went about it to figure it out? The errors in the console hinted something?
> 
> Also, in this fix we only add that weird extra-escaping just to match the signature, but keep the original payload when sending the message to Flex, right?

Exactly, we don't mess with the message we forward on, just the text we use for the input to the signature generation routine

PR #458, opened 2023-05-16
fix: set aws region env var to correct region from new ssm parameter

robert-bo-davis:
Updating CA hours to test the prod account, will do a patch release on Friday to correct the hours## Description

This moves to using an SSM parameter to set a helpline specific AWS_REGION env during build. The move to signature version v4 for S3 requires that the region for the client match the bucket's region.

I also went ahead and moved all s3 calls to use signature version v4.

My IDE lints everything I touch, so this is a bit hard to read. Turning off whitespace changes helps a bit.

The main IDE lint changes were:

- double quotes to single quotes in .yml files
- some newline nits
- a lot of whitespace cleanup.

The actual code changes were:

- Pass through full environment to custom_helpline everywhere it is called (chose this over a map because the shortEnv will go away with a move to the new SSM param layout)
- Update all s3 clients to use signature v4
- a slight tweak to the way deploy all staging matrix is defined.
- [load HELPLINE_AWS_REGION env](https://github.com/techmatters/serverless/pull/458/files#diff-86bb58e3276e06735d50ea8dd08dcc9d91e249a8039310d3ba8abe064e1dda9eR144) var from ssm and pass in via aws-region call.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/.github/actions/main-action/action.yml
+++ b/.github/actions/main-action/action.yml
@@ -22,16 +22,16 @@ inputs:
   auth-token:
     description: 'Twilio Token'
     required: true
-  workspace-sid:  
-      description: 'Twilio Workspace SID'
-      required: true
-  transfer-workflow-sid:  
+  workspace-sid:
+    description: 'Twilio Workspace SID'
+    required: true
+  transfer-workflow-sid:
     description: 'Chat Transfer Workflow SID'
     required: true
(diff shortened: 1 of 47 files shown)


stephenhand (approved):
LGTM!

PR #466, opened 2023-06-06
[Bug fix] Cold transfer stuck in "locked" state when user ends the chat during a transfer

GPaoloni:
This PR is related to https://github.com/techmatters/webchat/pull/256.

## Description
This PR fixes is part of the fix to https://tech-matters.atlassian.net/browse/CHI-1965https://tech-matters.atlassian.net/browse/CHI-1965.
This bug occurs because:
- When a new webchat task is created, we add the `taskSid` to the channel attributes, so we can easily find the corresponding task from by having the channel. This is used in the end chat functionality.
- When a transfer is initiated, a second task with the same channel sid is created. This means that the new "transfer task" (not yet accepted), because of the above behavior, was overriding the original `taskSid` in channel attributes.

In this PR, the task sids of all the tasks related to a channel are instead added to a list in the channel attributes. When the user ends the chat, we try to close all the open tasks related to the channel, moving them to the proper state, based on the task state at the time of ending the chat.

### Checklist
- [x] New tests added

### Verification steps
- Make sure this branch is deployed.
- Start a new local server for webchat using the branch mentioned above.
- Open two Flex sessions.
- Initiate a webchat contact and accept it in one of the Flex sessions.
- Transfer the webchat task from one Flex session to the other, but don't accept it in the second one.
- While the transfer is still in progress, end the chat from the webchat client.
- Confirm that the original task is in wrapup, and that the counselor can interact with the forms. Also the "transfer in progress task" should be removed from the second Flex session.

### DIFF
--- a/functions/endChat.ts
+++ b/functions/endChat.ts
@@ -78,8 +78,6 @@ const updateTaskAssignmentStatus = async (
   event: Body,
 ) => {
   const client = context.getTwilioClient();
-  let channelCleanupRequired = false;
-
   // Fetch the Task to 'cancel' or 'wrapup'
   const task = await client.taskrouter
     .workspaces(context.TWILIO_WORKSPACE_SID)
@@ -111,8 +109,7 @@ const updateTaskAssignmentStatus = async (
     case 'reserved':
     case 'pending': {
       await updateAssignmentStatus('canceled');
-      channelCleanupRequired = true;
-      break;
+      return true; // indicate that there's cleanup needed
     }
     case 'assigned': {
       await updateAssignmentStatus('wrapping');
@@ -121,7 +118,7 @@ const updateTaskAssignmentStatus = async (
     default:
   }
 
-  return channelCleanupRequired;
+  return false; // no cleanup needed
 };
 
 /**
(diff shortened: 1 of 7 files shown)


functions/endChat.ts
```diff
@@ -197,6 +185,17 @@ export const handler = TokenValidator(
         event,
       );
 
+      const members = await channel.members().list();
+      await Promise.all(
+        members.map((m) => {
```
stephenhand, line 190: Could we do an `async` function here?

functions/endChat.ts
```diff
@@ -77,51 +77,53 @@ const updateTaskAssignmentStatus = async (
...
-        body: endChatMessage,
-        from: 'Bot',
-        xTwilioWebhookEnabled: 'true',
-      });
-  }
-
-  // Update the task assignmentStatus
-  const updateAssignmentStatus = (assignmentStatus: TaskInstance['assignmentStatus']) =>
-    client.taskrouter
+  try {
```
stephenhand, line 80: I wonder if it might make more sense to just return the updated task from this function and let the caller figure out what the follow on action is based on it's state, rather than this function telling the caller what to do next, which seems to be outside its responsibility?

stephenhand (approved):
Couple of minor optional comments, looks good otherwise

PR #467, opened 2023-06-06
conference/updateParticipant: Added option to set muted

GPaoloni:
## Description
Adds one more valid update to `conference/updateParticipant`, allowing to set the `muted` attribute of conference participants.

### Checklist
- [x] New tests added

### Verification steps
- Call from the conference fronted and confirm it sets the value accordingly.


mythilytm (approved):
This helps for muting / unmuting behavior

PR #471, opened 2023-06-12
Bundle translations into a single file

murilovmachado:
## Description
We've reached another Twilio limitation: 50 private assets.
Because of that, we cannot deploy Serverless anymore.

In order to solve that, this PR bundles all translation assets into a single one at deploy time.


murilovmachado:
Closing this in favor of https://github.com/techmatters/serverless/pull/472

PR #472, opened 2023-06-12
Make translation files public assets

murilovmachado:
## Description
This PR makes all translation files to be public assets instead of private.
We're doing this because we've hit Twilio's limitation of 50 private assets.
Since these files are already public on GitHub we've decided that it won't have any harm to have them as public assets on Twilio.


stephenhand (approved):
Thanks Murilo!

PR #476, opened 2023-06-19
Load public assets with axios

murilovmachado:
## Description
Public assets cannot be fetched with `Runtime.getAssets()`.
This PR implements fetching the translation files (public) with `axios`.


acedeywin (approved):
LGTM

PR #480, opened 2023-06-26
Lex: Support to multiple languages

murilovmachado:
## Description
This PR adds support to multiple languages for Lex bots.

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-1792

### Verification steps
The easiest way to test: hardcode language on Studio flow. Then, test both languages.
Available languages for Aselo Dev: `en-US`, `es-CO`.


functions/webhooks/chatbotCallback.protected.ts
```diff
@@ -124,22 +117,35 @@ export const handler = async (
...
             .remove(),
           // Trigger the next step once the channel is released
-          nextAction,
+          nextAction(),
         ]);
 
         console.log('Channel unblocked and bot session deleted');
       }
 
+      // Why we need to send this message?
```
GPaoloni, line 142: The always want to send the Lex response to the channel, so the user can see it. The reason it is moved after all the other code is that `nextAction` now adds a webhook with a Studio trigger for pre survey, and the message being sent after will trigger the "second execution" of the flow so the channel triggers a task for Flex and all of that.

GPaoloni (approved):
LGTM, thanks!!

functions/webhooks/chatbotCallback.protected.ts
murilovmachado: Sorry, I've forgotten to remove this comment😅

I've noticed it later on. I just got confused initially when I was getting code from the Lex pivot stuff.
I'll remove the comment

PR #492, opened 2023-07-19
Adjust postSurveyInit and postSurveyListener to make them work with the new bot

GPaoloni:
## Description
This PR adapts the post survey flow to work properly with Lex, while also integrating it into the channel capture used for other kinds of Lex conversations.

The most relevant changes in this PR are:
- All function and logic around capturing chat channels is contained in `functions/channelCapture`.
- `channelCapture/channelCaptureHandlers` is the function where the core of this code lives. The code here is grouped like:
  - Trigger types and capture handlers, exposed via `handleChannelCapture`:
    Since both, pre survey and post survey are captures, we need to expose at least two kinds of triggers when capturing. There are several operations that both kinds of triggers share and are essential to the channel capture, which includes:
    - Creating a "control task", which will shut down the channel if the survey is never completed. This prevents channels going "stale".
    - Removing `studio` type webhooks, so the interactions with the bot don't execute the Studio Flow again.
    - Updating the chat channel with the information about the capture: bot, what action to perform after the survey is completed, where to store the bot memory, if there should be a flag set in the channel once the survey is complete.
    - Adding `chatbotCallback` function as a webhook invoked on message sent. This webhooks will exchange messages between the user and the Lex bot, and will "handle the channel release" once the survey is done (next section).
    The two different triggers are:
    - `withUserMessage` that takes a message already sent to the channel (usually from the child) and redirects it as it is to the Lex bot. From here user and chatbot exchange messages via webhook.
    - `withNextMessage` that takes a message that does not exists in the channel (this could be optional), sends it to the channel and only then creates the connection with the chatbot. This will leave the channel in a state where the next message will be the first to exchange with Lex, and from there the flow is the same as the above item.
  - Another important piece of the integration is the "release type", which refers to the action that should be performed when the survey ends and the channel is release. Here we also have two types of release handlers:
    - `triggerStudioFlow` release means once the survey is done, we'll create a new studio trigger and then send the last message from the bot. This results in the channel making it all the way into the Studio Flow. To prevent going into the same execution branch, we'll also set a "on release channel flag", e.g. `preSurveyComplete`, so we can make channels avoid survey if it was already completed. This is the release handler that we'll generally use for pre surveys, language selector bots or any other "information gathering survey" that needs the channel to be kept active after it completes.
    - `postSurveyComplete` is a unique kind of trigger, since this is always the last step in the life-cycle of a chat channel. Here we will save the post survey results in task attributes (insights) and in HRM.
- `channelCapture/captureChannelWithBot` function is just a wrapper around the above to expose the channel capture. It is intended to be called from Studio Flows primarily.
- `channelCapture/chatbotCallback` is the webhook used to exchange messages with the Lex bots. Here we'll "release the channel" when the conversation with the bot ends, saving the bot's memory in the channel attributes and calling the appropriate release handler.

Other changes in this PR include:
- Post survey code adapted to consume Lex memory.
- Post survey will be fired using the new Lex channel capture if `enable_lex` feature flag is on. Also post surveys will be triggered for custom channels now.
- In `taskrouterListeners/janitorListener`, prevent the janitor cleanup the channel for custom channels if the task is a capture control task. If it's pertinent to cleanup the channel for this case, that should be done in the `isCleanupBotCapture` case, not `isCleanupCustomChannel`.

### Checklist
- [x] Corresponding issue has been opened ([here](https://tech-matters.atlassian.net/browse/CHI-1795) and [here](https://tech-matters.atlassian.net/browse/CHI-1794))
- [ ] New tests added

### Verification steps
- Deploy this code to Aselo development
- Start a new webchat or Instagram contact
- Confirm that pre survey takes place
- After the contact is ended by the counselor, confirm that post survey takes place


functions/channelCapture/channelCaptureHandlers.private.ts
```diff
@@ -0,0 +1,531 @@
...
+  language: string; // (in Studio Flow, {{trigger.message.ChannelAttributes.pre_engagement_data.language | default: 'en-US'}} )
+  botSuffix: string;
+  triggerType: TriggerTypes;
+  releaseType: ReleaseTypes;
+  studioFlowSid?: string; // The Studio Flow sid. Needed to trigger an API type execution once the channel is released. (in Studio Flow, flow.flow_sid)
+  memoryAttribute?: string; // Where in the channel attributes we want to save the bot's memory (allows usage of multiple bots in same channel)
+  releaseFlag?: string; // The flag we want to set true in the channel attributes when the channel is released
+  additionControlTaskAttributes?: string; // Optional attributes to include in the control task, in the string representation of a JSON
+  controlTaskTTL?: number;
+};
```
murilovmachado, line 251: I think you have this same time defined somewhere else, right?
Can't we local `import` types in this repo? It may work, but I'm not sure.

functions/postSurveyComplete.protected.ts
```diff
@@ -71,14 +73,15 @@ const saveSurveyInInsights = async (
...
   await surveyTask.update({ attributes: JSON.stringify(finalAttributes) });
 };
 
 const saveSurveyInHRM = async (
   postSurveyConfigJson: OneToManyConfigSpec[],
-  memory: BotMemory,
+  memory: AutopilotMemory,
   surveyTask: TaskInstance,
   surveyTaskAttributes: any,
   hrmBaseUrl: string,
```
murilovmachado, line 87: This means we continue saving memory (in hrm and insights) in the Autopilot format, right?

functions/postSurveyInit.ts
```diff
@@ -152,7 +190,17 @@ export const handler = TokenValidator(
       if (channelSid === undefined) return resolve(error400('channelSid'));
       if (taskSid === undefined) return resolve(error400('taskSid'));
 
-      await postSurveyInitHandler(context, { channelSid, taskSid, taskLanguage });
+      const result = await postSurveyInitHandler(context, {
+        channelSid,
+        taskSid,
+        taskLanguage: taskLanguage || 'en-US',
+      });
```
murilovmachado, line 197: Just a question: Would it make sense to move the fallback `en-US` to Studio? I think this fallback wouldn't be very useful for helplines that don't operate in English, for example.

murilovmachado (approved):
Looks excelent, Gian! Just a few minor questions 🚀

functions/channelCapture/channelCaptureHandlers.private.ts
```diff
@@ -0,0 +1,531 @@
+/* eslint-disable global-require */
```
stephenhand, line 1: I like this approach, we can extend it to be a full multicast event system in the future if we ever need to.

stephenhand (approved):
Initial look over LGTM, will probably revisit another time or 2 just to make sure I fully understand it but you're good to merge for me

functions/postSurveyComplete.protected.ts
GPaoloni: In particular this part right here means that `saveSurveyInHRM` function in `functions/postSurveyComplete.protected.ts` will use Autopilot format, but I duplicated the same functions for HRM and Insights in `/channelCapture/channelCaptureHandlers.private.ts`, since the first ones will be removed soon (once we are all in Lex).

functions/postSurveyInit.ts
GPaoloni: You are right. In that case we should "fail" if language is not provided, which is not a bad thing at all.

PR #494, opened 2023-07-26
[Bug fix] Duplicated Line contact created when transfer rejected

GPaoloni:
## Description
This PR fixes the bug where, after rejecting a line transfer (or any custom channel really), the next message will result in a duplicated contact. 
The were two causes for this bug:
- [Our implementation of the "chat channel janitor"](https://github.com/techmatters/serverless/blob/master/functions/taskrouterListeners/janitorListener.private.ts) was triggered when the transferred task was rejected (hence canceled), resulting in the chat channel being deactivated. The next message will result in an entire new Studio Flow execution which will ultimately be a new task in Flex.
- Twilio's "Channel Janitor" is enabled for all custom channels in Thai staging, which will do the same as above but on every task being completed/canceled, as this is the default behavior. 
When a task is transferred:
  - A new task will be created to assign to the target counsellor. This task will have `channelSid` attribute set to the original channel in use for the contact. 
  - The original task attributes are modified, among those, the `channelSid` is set to a dummy value. Because of this, when the transfer is accepted, and the original task is canceled, the channel is not deactivated by the Twilio's janitor, as it does not points to the `channelSid` in use.
The logic in this second bullet point was not true for the newly created task in the transfer process, hence when this one was canceled, the Twilio's janitor will have access to the original `channelSid` in use, and deactivates it.
The fix for this is as simple as setting the `channelSid` attribute to a dummy value in the transferred task, when it's rejected.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-1959)
- [x] New tests added

### Verification steps
Follow the repro steps in the linked Jira ticket and confirm the issue is not occuring anymore (this code is deployed in Thai staging).

### DIFF
--- a/.github/workflows/thailand_staging.yml
+++ b/.github/workflows/thailand_staging.yml
@@ -26,3 +26,4 @@ jobs:
       environment_code: 'STG'
       environment: 'staging'
       changelog: 'Deploying individual account'
+      send-slack-message: false
--- a/.github/workflows/thailand_staging.yml
+++ b/.github/workflows/thailand_staging.yml
@@ -26,4 +26,3 @@ jobs:
       environment_code: 'STG'
       environment: 'staging'
       changelog: 'Deploying individual account'
-      send-slack-message: false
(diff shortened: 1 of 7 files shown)


acedeywin (approved):
@GPaoloni  - This works perfectly. LGTM

PR #495, opened 2023-08-02
AS Staging Setup

sinehome:
Setting up staging environment


.github/workflows/deploy-all-staging-accounts.yml
```diff
@@ -28,28 +28,8 @@ jobs:
...
-            'ZA',
-            'CO',
-            'TH',
-            'UK',
-            'HU',
-            'ZM',
-            'ZW',
-            'PL',
-            'MT',
-            'NZ',
```
janorivera, line 51: NZ is missing in the update, and the format should be the same. Be sure to add the comma at the end or else the lint test will fail.

janorivera (changes requested):

janorivera (approved):
lgtm!

PR #503, opened 2023-08-28
CHI-1260: Deploy all prod workflow

stephenhand:
## Description

* Add a generic 'deploy multiple helplines for multiple environments workflow'
* Refactor 'Deploy All Staging Accounts' workflow to use above workflow
* Add a 'Deploy All Production Accounts' workflow using the same workflow
* Updated deploy workflow to only need the environment specifying once
* Updated single helpline deploy to use multiple deploy as well

### Checklist

### Related Issues
CHI-1260

### Verification steps

* Test the new 'deploy all prod' workflow & regression test the 'deploy all staging' and single helpline deploy workflows (particularly dev)


stephenhand:
@robert-bo-davis @GPaoloni FYI

.github/workflows/config/production-helplines.json
```diff
@@ -0,0 +1,3 @@
+[
+  "AS"
+]
```
murilovmachado, line 3: You have only `AS` here for testing, but eventually, we'll list them all, correct?

murilovmachado (approved):
Looks good to me. I'm going to borrow the same idea for the `flex-plugins` repo

.github/workflows/config/production-helplines.json
stephenhand: Yes, sorry, should have explained in the PR it's the old chicken & egg problem, shouldn't merge the PR before it's tested but can't test the new workflow without it being merged

stephenhand: Yes, sorry should have noted that on the PR

The old 'chicken & egg' problem that we shouldn't merge before we test but the new workflow can't be tested before it's merged

PR #505, opened 2023-08-29
CHI-1260: Infer staging accounts to deploy

stephenhand:
## Description

* Create a local action to read hewlpline short codes based on entries in the SSM parameter store
* Automatically infer which staging accounts to deploy to based on the existence of the SSM parameters created as part of the account setup, removing the need for a separately maintained account list in the repo

### Checklist
- N/A New tests added

### Related Issues
CHI-1260

### Verification steps

Run the 'Deploy All Staging Accounts' workflow from this branch & ensure the expected accounts are successfully deployed


robert-bo-davis (approved):
This brings me joy. Thanks!

One question is:  I think this action will probably end up being used in all of our repos. Do we want to keep it in serverless or would one of the others make more sense to become our "all shared action logic goes here" repo?

stephenhand:
@robert-bo-davis  FYI ChatGPT almost got it right, but not almost right enough to prevent me having to wrangle with bash nested interpolations 🤮 😄

PR #511, opened 2023-09-07
[Bug fix] Make channel capture cleanup function compatible with channel not captured

GPaoloni:
## Description
This PR fixes some issues found in `/channelCapture/chatbotCallbackCleanup` function, caused because of the assumption that the channel is successfully captured when it may not.

The changes included in this PR are:
- `/channelCapture/channelCaptureHandlers` continues to bubble the error (if any) from the Lex response, but only after the channel attributes have been updated, since those attributes are required in the cleanup function.
- Lex client's `deleteSession` function is wrapped in a Result to avoid throwing, as that can cause the cleanup to error when the Lex session does not exists (due to a channel capture error). With this change we prevent the promise to reject. If we ever need to "consume the error", that can be done by just evaluating the returned result, but right now we don't care about it.
- Remove remaining `console.error`s to avoid paging people :)

### Verification steps
If @janorivera approves this PR, it means he already tested the behavior :P

### DIFF
--- a/functions/channelCapture/channelCaptureHandlers.private.ts
+++ b/functions/channelCapture/channelCaptureHandlers.private.ts
@@ -160,10 +160,6 @@ const triggerWithUserMessage = async (
     inputText,
   });
 
-  if (lexResult.status === 'failure') {
-    throw lexResult.error;
-  }
-
   const chatbotCallbackWebhook = await channel.webhooks().create({
     type: 'webhook',
     configuration: {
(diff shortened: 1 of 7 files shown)


janorivera:
I have tested this and it works.

1. Tested with a NOT BUILT bot
The task goes to Task Router and to the respective task queue. The FB channel doesn't get "stuck" as before.

2. Tested with a BUILT bot
The bot flow is executed normally and prepopulate works as expected.

janorivera (approved):

PR #522, opened 2023-10-05
Prevent sendStudioMessage to trigger another studio execution

murilovmachado:
## Description
This PR fixes an issue where `sendStudioMessage` was triggering another Studio execution.
This is fixed by removing the webhook of type `studio` before sending a message.

### Related Issues
Fixes https://tech-matters.atlassian.net/browse/CHI-2170

### Verification steps
Alejandro already checked that this is not happening anymore with this code.
If you want to double-check, you can use Instagram to contact Aselo Development.


acedeywin (approved):
LGTM

PR #524, opened 2023-10-27
Chi 2213 migrate e2e account autopilot lex

acedeywin:
## Description

Refactor how to set helpline code

- Since we can't provision a lex chatbot with e2e as the helpline code we will have to change the helpline code for the E2E account to "ee" on serverless

### Checklist
- [ ] New tests added

### Related Issues
Fixes [CHI-2213](https://tech-matters.atlassian.net/browse/CHI-2213)

### Verification steps

Deploy to all helpline and make sure they were all successful

[CHI-2213]: https://tech-matters.atlassian.net/browse/CHI-2213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/deploy-multiple-accounts.yml
+++ b/.github/workflows/deploy-multiple-accounts.yml
@@ -124,19 +124,24 @@ jobs:
           ssm_parameter: ${{env.TWILIO_SSM_PREFIX}}_POST_SURVEY_BOT_CHAT_URL
           env_variable_name: POST_SURVEY_BOT_CHAT_URL
 
-      - name: Check matrix.helpline-code and set HELPLINE_CODE
+      - name: Set Helpline Code for E2E Account
+        if: matrix.helpline_code == 'E2E'
+        uses: marvinpinto/action-inject-ssm-secrets@latest
+        with:
+          ssm_parameter: ${{env.TWILIO_SSM_PREFIX}}_HELPLINE_CODE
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-multiple-accounts.yml
```diff
@@ -123,6 +123,28 @@ jobs:
         with:
           ssm_parameter: ${{env.TWILIO_SSM_PREFIX}}_POST_SURVEY_BOT_CHAT_URL
           env_variable_name: POST_SURVEY_BOT_CHAT_URL
+
+      # Since we can't provision a lex chat bot with e2e as the helpline code
+      # we will have to change the helpline code for the E2E account to "ee" on serverless
+      - name: Set Helpline Code for E2E Account
```
stephenhand, line 129: We shouldn't have helpline specific exceptions in the core deploy workflow.

We should either:

- make this logic generic, like using the HELPLINE_CODE parameter if it is set for the helpline, or fall back to the existing logic if not

OR

- Add the _HELPLINE_CODE lookup specifically in a custom action for the E2E account - custom actions are run after the core setup so .env lines added there will override the core ones

I don't really have a preference for either

PR #550, opened 2023-12-15
Bg fix issue with e2e

acedeywin:
## Description

- This PR looks into handling a situation where the helpline code contains a number. It would match all digits (0-9) and replace them with no space since Lex doesn't accept numbers.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/functions/channelCapture/channelCaptureHandlers.private.ts
+++ b/functions/channelCapture/channelCaptureHandlers.private.ts
@@ -360,7 +360,13 @@ export const handleChannelCapture = async (
   ]);
 
   const { ENVIRONMENT, HELPLINE_CODE } = context;
-  const languageSanitized = language.replace('-', '_'); // Lex doesn't accept '-'
+  let languageSanitized = language.replace('-', '_'); // Lex doesn't accept '-'
+
+  //This is used to match all digits (0-9) and replace them with no space since Lex doesn't accept numbers
+  if(/\d/.test(languageSanitized)){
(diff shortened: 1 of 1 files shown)


functions/channelCapture/channelCaptureHandlers.private.ts
```diff
@@ -360,7 +360,13 @@ export const handleChannelCapture = async (
   ]);
 
   const { ENVIRONMENT, HELPLINE_CODE } = context;
-  const languageSanitized = language.replace('-', '_'); // Lex doesn't accept '-'
+  let languageSanitized = language.replace('-', '_'); // Lex doesn't accept '-'
+
+  // This is used to match all digits (0-9) and replace them with no space since Lex doesn't accept numbers
+  if (/\d/.test(languageSanitized)) {
```
robert-bo-davis, line 366: Do we need an if wrapper? Will replace return the same string if the regex isn't found?

robert-bo-davis:
Were we also going to go ahead and revert the github action changes that tried to set the HELPLINE_CODE env var to `ee`?

stephenhand (approved):
Lgtm except for @robert-bo-davis comment

functions/channelCapture/channelCaptureHandlers.private.ts
acedeywin: @robert-bo-davis  - If there is no number the if block is ignored. So the initial `languageSanitized` will return its original value.

PR #561, opened 2024-01-09
fix: safeguards chatbotCallbackCleanup

GPaoloni:
## Description
This PR adds safe-guards to `chatbotCallbackCleanup` function, in case that `capturedChannelAttributes` is not present in channel attributes. This is in order to allow the cleanup to succeed even in those cases where the capture function fails before even attempting to create this property.

image

### Checklist
- [ ] New tests added

### Verification steps
- Deploy this code to Development.
- Update Messaging Studio Flow, passing an invalid parameter to the "capture channel with bot" widget.
- Start a new web contact.
- Confirm that even when the capture fails, the chat is not blocked (the capture is "cleaned-up").

### DIFF
--- a/functions/channelCapture/chatbotCallbackCleanup.protected.ts
+++ b/functions/channelCapture/chatbotCallbackCleanup.protected.ts
@@ -69,10 +69,10 @@ export const chatbotCallbackCleanup = async ({
 
   const releasedChannelAttributes = {
     ...omit(channelAttributes, ['capturedChannelAttributes']),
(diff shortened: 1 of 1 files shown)


GPaoloni:
@janorivera I'm interested in you testing this PR if possible.

janorivera:
OK I've tested this and it works. If the chatbot function fails, the clean up function is successfully executed.
image

janorivera (approved):
lgtm!

PR #618, opened 2024-05-20
CHI-2685: Custom channel conversations

stephenhand:
## Description

Added code to custom channel functions to support conversations. I haven't gone crazy doing nice structures or making everything super DRY for Programmable Chat vs Conversations because we expect Programmable Chat support to go away ASAP once we complete the conversations migration

I have marked all the programmable chat specific code as @deprecated in the JS Doc in the meantime

Additional work:

Some more serverless functions were branched to use conversations APIs if we are using conversations, we were implicitly relying on backwards compatibility for several processes in built in channels, but this did not work immediately for custom channel conversations, so we moved them to use the conversations APIs correctly

_Once we are fully migrated we should do a sweep of the code for any chat service APIs, we shouldn't need them anywhere_

The 'conversation updated' and 'conversation state updated' events do not fire their webhooks if the conversation state is updated by flex, rather than directly via the API. This prevented us using a webhook to remove closed conversations from the userChannelMap, resulting in stuck conversations. To work around this we now have a 'triple check' - the map entry is removed as part of the task.completed event handler, we also check for closed conversations in the map when deciding if a new conversation is required. Finally, we still bind to the onConversationUpdated ands onConversationStateUpdated events, just in case we start changing this state via the API.

The `channelType` task attribute isn't set by the studio flow correctly. We are still discussing the details with Twilio support but in the meantime we are using a second property 'customChannelType'. Flex will need to be updated to consume this property

The Line 'send message' implementation has been migrated to use fetch rather than axios

A ConversationSid and ChannelSid type have been introduced

TODO: unit test coverage (can be follow on PR)

### Checklist
- [ ] New tests added
- [X] Feature flags / configuration added

### Other Related Issues

None

### Verification steps

Full regression test on Line channel

### AFTER YOU MERGE


functions/channelCapture/chatbotCallback.protected.ts
```diff
@@ -42,14 +47,15 @@ type EnvVars = AWSCredentials & {
   SURVEY_WORKFLOW_SID: string;
 };
 
-export type Body = Partial<WebhookEvent> & {};
+export type Body = Partial<ConversationWebhookEvent & ProgrammableChatWebhookEvent> & {};
```
GPaoloni, line 50:
```suggestion
export type Body = ConversationWebhookEvent | ProgrammableChatWebhookEvent;
```
does not works here?

functions/taskrouterListeners/createContactListener.private.ts
```diff
@@ -55,6 +55,10 @@ export const handleEvent = async (context: Context<EnvVars>, event: EventFields)
 
     const taskAttributes = JSON.parse(taskAttributesString);
 
+    Object.entries(taskAttributes).forEach(([key, value]) => {
+      console.log('Task attribute:', key, value);
+    });
+
```
GPaoloni, line 61: Intended? :P

GPaoloni (approved):
Code looks good to me!

> I haven't gone crazy doing nice structures or making everything super DRY for Programmable Chat vs Conversations because we expect Programmable Chat support to go away ASAP once we complete the conversations migration

Reduced cognitive load is also appreciated here :P

> The 'conversation updated' and 'conversation state updated' events do not fire their webhooks if the conversation state is updated by flex, rather than directly via the API. This prevented us using a webhook to remove closed conversations from the userChannelMap, resulting in stuck conversations. To work around this we now have a 'triple check' - the map entry is removed as part of the task.completed event handler, we also check for closed conversations in the map when deciding if a new conversation is required. Finally, we still bind to the onConversationUpdated ands onConversationStateUpdated events, just in case we start changing this state via the API.

About "the map entry is removed as part of the task.completed event handler", I think that entirely eliminates the possibility of post-survey chatbot interactions, since we loose the information of "to which channel route messages from a given identifier".. Maybe this is sorted somehow and I'm missing it?
Edit: I wrote the above before the review. What you mean by "the map entry is removed as part of the task.completed event handler" is actually the code of `functions/taskrouterListeners/janitorListener.private.ts`?

functions/channelCapture/chatbotCallback.protected.ts
stephenhand: Not without rewriting the validation logic, which I can't be bothered doing since Programmable Chat support will be going away

functions/helpers/chatChannelJanitor.private.ts
```diff
@@ -24,14 +24,47 @@
...
-}
+export type Event =
+  | {
+      channelSid?: ChatChannelSid;
+      conversationSid: ConversationSid;
+    }
+  | {
+      channelSid: ChatChannelSid;
+      conversationSid?: ConversationSid;
+    };
```
murilovmachado, line 37: I've spent some time trying to spot the difference between the two types xD

functions/helpers/customChannels/customChannelToFlex.private.ts
```diff
@@ -16,7 +16,14 @@
 
 import { Context } from '@twilio-labs/serverless-runtime-types/types';
 
+export type ConversationSid = `CH${string}`;
+export type ChatChannelSid = `CH${string}`;
+
+const CONVERSATION_CLOSE_TIMEOUT = 'P3D';
```
murilovmachado, line 22: Can you add a comment on what `P3D` means?

functions/webhooks/line/FlexToLine.protected.ts
```diff
@@ -54,84 +56,94 @@ const sendLineMessage =
...
-          }`,
-        );
-      }
-      throw error;
-    }
+    const response = await fetch(LINE_SEND_MESSAGE_URL, {
+      method: 'post',
+      body: JSON.stringify(payload),
+      headers,
+    });
```
murilovmachado, line 69: Using `fetch`, nice!

murilovmachado (approved):
Great Job!

functions/helpers/customChannels/customChannelToFlex.private.ts
stephenhand: I will link to the ISO standard wiki page, good call

PR #635, opened 2024-07-03
CHI-2818: Add support for backend pulling

stephenhand:
## Description

This is an attempt to fix the manual pulling issue where transfers leave the worker with too high a chat capacity

The theory is that the flex event we rely on is not fired when the task is completed outside of flex, so we are moving he whole function out of flex and into task router events, where events should be consistent and respected

* Adds a listener to reduce chat capacity of the worker once the task is completed (replicating flex functionality)
* Adds a 'pullTask' endpoint. This will increase the chat capacity of the worker, poll the reservations for this worker for 5 seconds to see if a new one is added - if it is a task is considered to have successfully been 'pulled' - returns a 200 and the task SID. If there is no new reservation seen in the timeout period, a 404 is returned and the chat capacity is reduced again
* Adds flag to control this behaviour - if not set it assumes manual pulling is managed flex side.

This also needs the changes in this Flex PR to work: TBD

### Checklist
- [ ] New tests added
- [X] Feature flags / configuration added

### Other Related Issues

None

### AFTER YOU MERGE

### DIFF
--- a/functions/channelCapture/captureChannelWithBot.protected.ts
+++ b/functions/channelCapture/captureChannelWithBot.protected.ts
@@ -71,7 +71,7 @@ export const handler = async (
 
     const isConversation =
       (typeof isConversationValue === 'string' && isConversationValue === 'true') ||
-      Boolean(isConversationValue);
+      (typeof isConversationValue === 'boolean' && isConversationValue);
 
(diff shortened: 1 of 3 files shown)


murilovmachado (approved):
Looks good to me!

GPaoloni (approved):

PR #637, opened 2024-07-05
Chi 2811 support media transcripts

acedeywin:
## Description

- New feature: This is the API call implementation for the Twilio conversations media resource API used to get the temporary URL for media displayed on the transcript(Here is [Flex-plugin PR](https://github.com/techmatters/flex-plugins/pull/2377) )

- Next Course of Action: To implement the unit test for the getMediaUrl function.

### Checklist
- [ ] Corresponding issue has been opened
- [ ] New tests added
- [ ] Feature flags / configuration added

### Other Related Issues

None

[CHI-2811](https://tech-matters.atlassian.net/browse/CHI-2811)

### AFTER YOU MERGE

1. Cut a release tag using the GitHub workflow. Wait for it to complete and the notification to be posted in the #aselo-deploys Slack channel.
2. Comment on the ticket with the release tag version AND any additional instructions required to configure an environment to test the changes.
3. Only then move the ticket into the QA column in JIRA

You are responsible for ensuring the above steps are completed. If you move a ticket into QA without advising what version to test, the QA team will assume the latest tag has the changes. If it does not, the following confusion is on you! :-P

[CHI-2788]: https://tech-matters.atlassian.net/browse/CHI-2788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CHI-2811]: https://tech-matters.atlassian.net/browse/CHI-2811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #638, opened 2024-07-08
CHI-2788: Add unit test for getmediaUrl function

acedeywin:
## Description

- Test: this is the unit test for the [getMediaUrl function](https://github.com/techmatters/serverless/pull/637) 

### Checklist
- [x] New tests added

### Other Related Issues

None

### AFTER YOU MERGE


tests/getMediaUrl.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import helpers, { MockedResponse } from './helpers';
+
+jest.mock('@tech-matters/serverless-helpers', () => ({
+  ...jest.requireActual('@tech-matters/serverless-helpers'),
+  functionValidator: (handlerFn: any) => handlerFn,
+}));
+
+jest.mock('axios', () => ({
+  request: jest.fn(),
+}));
```
murilovmachado, line 30: Stephen, I missed the original Serverless code, but I think we should not be using `axios` anymore. We're moving away from `axios` to use node's `fetch`.

This test is very good that it made me realize that we should change the Serverless code to use `fetch`. Can you do that as part of this PR?

tests/getMediaUrl.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    const event1: Event = {
+      serviceSid: undefined,
+      mediaSid: 'MIxxxxxxxxxxIOL',
+      request: { cookies: {}, headers: {} },
+    };
+
+    const emptyEvent = {
+      mediaSid: undefined,
+      request: { cookies: {}, headers: {} },
+    };
```
murilovmachado, line 68: The test description says `if value is undefined`. Which value is this? the `serviceSid` or `mediaSid`? Can you specify this in the test description?

tests/getMediaUrl.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    await getMediaUrl(baseContext, event1, callback);
+    await getMediaUrl(baseContext, emptyEvent, callback);
+  });
+
+  test('Should return status 500 if data is undefined', async () => {
+    const event: Event = {
+      serviceSid: 'ISxxxxxxxxxxxxxxAWX',
+      mediaSid: 'MIxxxxxxxxxxIOL',
+      request: { cookies: {}, headers: {} },
+    };
```
murilovmachado, line 85: Same here: `if data is undefined`. Can you change the words to tell which `data` is it?

murilovmachado (changes requested):
Let's try not to use `axios` anymore. I know the original code got merged with `axios`, but we should be using `fetch` instead

tests/getMediaUrl.test.ts
acedeywin: @murilovmachado - I have updated the code and test as requested.

functions/getMediaUrl.ts
```diff
@@ -18,7 +18,7 @@
 /* eslint-disable import/no-dynamic-require */
 import '@twilio-labs/serverless-runtime-types';
 import { Context, ServerlessCallback } from '@twilio-labs/serverless-runtime-types/types';
-import axios from 'axios';
+import fetch from 'node-fetch';
```
murilovmachado, line 21: Great, thank you for making the change

tsconfig.eslint.json
```diff
@@ -9,5 +9,6 @@
   "include": [
     "functions",
     "tests",
+    "jest.setup.ts"
```
murilovmachado, line 12: Do we need this?

murilovmachado (approved):
Looks good to me. Just one question on `tsconfig.eslint.json`

tsconfig.eslint.json
acedeywin: Not at all. Let me remove it.

PR #648, opened 2024-07-18
CHI-2858: Tolerate failure to transition some agent participants into the target status

stephenhand:
## Description

Work around issue where previous agents in transferred tasks still appear as participants in the interaction but cannot have their status transitioned (presumably because they are no longer active in the underlying conversation?)

See code comment for specifics

Also, there is a fix for another issue spotted with transferring custom channel conversations. The code intended to remove the sending agent from a conversation in a transfer actually removes the caller in the case of custom channels, because the logic to locate the agent conversation participant didn't work for custom channels

I've changed the approach to identifying the interaction channel participant (by the type property being set as 'agent') rather than the conversation participant, storing THAT sid on the task and setting the status on the participant to closed on a successful transfer.

Because the logic to transition interaction channel participant status is called from a listener as well as a public function, I've moved the core logic into a private function where it can be invoked from both places

### Checklist
- N/A New tests added
- N/A Feature flags / configuration added

### Other Related Issues

None

### Verification steps

See ticket - also check non transferred tasks still work ok 

### AFTER YOU MERGE


murilovmachado (approved):
🚀 🚀

PR #649, opened 2024-07-19
CHI-2790: Alert service user for unsupported media

acedeywin:
## Description

- New feature: This PR implements a webhook that would send a message to callers whenever they send unsupported media. This can also cover the custom channel [ticket](https://tech-matters.atlassian.net/browse/CHI-2841) since the message sent out is generic as we cannot determine the media type for the unsupported message. 

- Next Action: To implement a unit test for the newly implemented `serviceConversationListener` webhook.

### Checklist
- [ ] New tests added

### Other Related Issues

None

[CHI-2790](https://tech-matters.atlassian.net/browse/CHI-2790)

### Verification steps

- Start a conversation (WhatsApp, Line, Telegram, etc)
- Send an unsupported media file (location in WhatsApp, Stick or image in Line/Telegram)
- Verify that you get an automatic response notifying you about the reaction not delivered

### AFTER YOU MERGE

[CHI-2790]: https://tech-matters.atlassian.net/browse/CHI-2790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


functions/webhooks/serviceConversationListener.protected.ts
```diff
@@ -0,0 +1,125 @@
...
+  const response = responseWithCors();
+  const resolve = bindResolve(callback)(response);
+  try {
+    const { EventType, ConversationSid, MessageSid, ParticipantSid, Body } = event;
+
+    let messageAuthor: string | undefined;
+
+    // check if it's the onMessageAdded event
+    if (EventType === 'onMessageAdded') {
+      const conversationMessage = await context
```
stephenhand, line 88: Can we move the logic for sending errors for unsupported media into a separate private function.  The top level webhook handler should just be a general conversation webhook handler that delegates to other functions for specific purposes. If there is a problem with the error messages in the future it will be much easier to find the relevant code if it is in a function module called something like `sendErrorMessageForUnsupportedMedia.private.ts` than if it just somewhere in the generic `serviceConversationListener.protected.ts`

functions/webhooks/serviceConversationListener.protected.ts
```diff
@@ -0,0 +1,125 @@
...
+        .conversations.v1.conversations(ConversationSid)
+        .messages(MessageSid)
+        .fetch();
+
+      const participantsList = await context
+        .getTwilioClient()
+        .conversations.v1.conversations(ConversationSid)
+        .participants.list();
+
+      participantsList.forEach((participant) => {
```
stephenhand, line 99: In the custom channel logic, we check the `participantSid` attribute of the conversation (under the `attributes` property, not the participant list). It would be safer to replicate the logic we know is working there than doing something new here

stephenhand (commented):
Couple of comments

functions/webhooks/serviceConversationListener.protected.ts
stephenhand, line 88: We shouldn't need to pull the message from the API, all the properties of the message should come through in the webhook already

functions/webhooks/serviceConversationListener.protected.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (participant.sid !== conversationMessage.participantSid) {
+          // The message author has to be the participant receiving the message (counsellor) not the sender
+          messageAuthor = participant.identity;
+        }
+      });
+
+      if (
+        ParticipantSid === conversationMessage.participantSid &&
```
stephenhand, line 107: This check is pointless. ParticipantSid in the webhook is the message participantSid, so it is comparing something to itself.

In the custom channel logic, we check the participantSid attribute of the conversation (under the attributes property of the conversation, not the participant list) - best to copy that as we know it works.

functions/webhooks/serviceConversationListener.protected.ts
```diff
@@ -0,0 +1,125 @@
...
+
+      const participantsList = await context
+        .getTwilioClient()
+        .conversations.v1.conversations(ConversationSid)
+        .participants.list();
+
+      participantsList.forEach((participant) => {
+        if (participant.sid !== conversationMessage.participantSid) {
+          // The message author has to be the participant receiving the message (counsellor) not the sender
+          messageAuthor = participant.identity;
```
stephenhand, line 102: Again the Author comes through as a webhook event property, we don't need to go digging through the API for it - see the onMessageAdded handler  in the flexToCustomChannel.private.ts function

acedeywin:
@stephenhand - All feedback has been implemented.

functions/helpers/sendErrorMessageForUnsupportedMedia.private.ts
```diff
@@ -0,0 +1,91 @@
...
+    ? `${differenceInSeconds} ${unitCheck(differenceInSeconds, 'second')}`
+    : `${differenceInMinutes} ${unitCheck(differenceInMinutes, 'minute')}`;
+};
+
+export const sendErrorMessageForUnsupportedMedia = async (context: Context, event: Event) => {
+  const { EventType, Body, Media, ConversationSid, DateCreated } = event;
+
+  if (EventType === 'onMessageAdded' && !Body && !Media) {
+    const messageTime = await getTimeDifference(DateCreated);
+    const messageText = `Sorry, your reaction sent ${messageTime} ago could not be delivered. Please send another message.`;
```
stephenhand, line 81: The message isn't necessarily a reaction, should probably say 'message'. Also 'Please send another message' indicates that resending the message might work in the future. We want to make it clear that what they are attempting is unsupported and will never work

functions/helpers/sendErrorMessageForUnsupportedMedia.private.ts
```diff
@@ -0,0 +1,91 @@
...
+  // Determine whether to return the difference in seconds or minutes
+  return differenceInSeconds < 60
+    ? `${differenceInSeconds} ${unitCheck(differenceInSeconds, 'second')}`
+    : `${differenceInMinutes} ${unitCheck(differenceInMinutes, 'minute')}`;
+};
+
+export const sendErrorMessageForUnsupportedMedia = async (context: Context, event: Event) => {
+  const { EventType, Body, Media, ConversationSid, DateCreated } = event;
+
+  if (EventType === 'onMessageAdded' && !Body && !Media) {
```
stephenhand, line 79: It looks like you are just detecting an empty message here - could you comment the code with why this is the approach rather than looking for something on the message specifically indicating an error?

stephenhand (approved):
Minor comments

Could we also test this with custom channels - hopefully it will 'just work' there too

PR #657, opened 2024-07-25
CHI-2818 queue transfer capacity workaround

stephenhand:
## Description

Adds explicit chat capacity reductions for worker with original task after a transfer. THis prevents a race condition where they can be assigned a new task before their chat capacity is decreased due to completing the task

Also adds an 'increaseUntilCapacityAvailable' mode for adjustChatCapactity that will keep increasing chat capacity until capacity becomes available or the max capacity is reached. I was concerened that 'double decreases' could leave a worker with too low a capacity to pull tasks in some scenarios so this works around that

### Checklist
- [ ] New tests added

### Other Related Issues

None

### Verification steps

See ticket - ensure worker and queue transfer scenarios are working

Also, ensure that if a worker picks up 3 tasks, then transfers 1 away, they can still pull yet another from the queue.

### AFTER YOU MERGE

### DIFF
--- a/functions/adjustChatCapacity.ts
+++ b/functions/adjustChatCapacity.ts
@@ -24,6 +24,7 @@ import {
   send,
   functionValidator as TokenValidator,
 } from '@tech-matters/serverless-helpers';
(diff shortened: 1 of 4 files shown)


murilovmachado (approved):
Looks good to me

PR #664, opened 2024-08-12
CHI-2818: Change capacity adjustment approach

stephenhand:
## Description

Instead of incrementing by 1 and decrementing by 1, when we need extra capacity we increase configured capacity until the API reports we have some capacity available (which will effectively be the current number of tasks the worker has going +1). Then when any task (or reservation?) is accepted, we reset the configured capacity to 1. This means that the worker will be 'over capacity' whenever they have manually pulled tasks or accepted transfers in addition to other tasks. I can't see that being 'over capacity' has any ill effects, and it means that there will never be circumstances where a worker incorrectly gets assigned new tasks because their capacity is higher than it ought to be.

So what will happen is:

1. Before manually pulling / being offered a transfer, the workers capacity should always be 1.
2. When the action happens, their capacity gets bumped to however many tasks they currently have +1
3. As soon as the worker accepts or rejects, their capacity is reduced back to 1 - they are 'over capacity'.
4. When they complete their task, their capacity is already too low for new tasks so the race condition isn't a problem, we don't even change capacity on task completion in this model

I've tested this when the worker pulls a task manually then gets offered a transfer before they accept the task they pulled, changing the chat capacity doesn't appear to revoke reservations that are already pending, so this case is covered.

### Checklist
- [ ] New tests added
- [X] Feature flags / configuration added (covered by `enable_backend_manual_pulling`

### Other Related Issues

None

### Verification steps

Test issues raised in original ticket and thorough regression of manual pulling, transfers and their intersection

### AFTER YOU MERGE

### DIFF
--- a/functions/adjustChatCapacity.ts
+++ b/functions/adjustChatCapacity.ts
@@ -32,7 +32,7 @@ type EnvVars = {
 
 export type Body = {
   workerSid?: string;
-  adjustment?: 'increase' | 'decrease' | 'increaseUntilCapacityAvailable';
+  adjustment?: 'increase' | 'decrease' | 'increaseUntilCapacityAvailable' | 'setTo1';
   request: { cookies: {}; headers: {} };
 };
 
@@ -118,6 +118,17 @@ export const adjustChatCapacity = async (
 
     // If configuredCapacity is already 1, send status 200 to avoid error on client side
     return { status: 200, message: 'Successfully decreased channel capacity' };
+  }
+
+  if (body.adjustment === 'setTo1') {
+    if (channel.configuredCapacity !== 1) {
+      await channel.update({ capacity: channel.configuredCapacity - 1 });
+      // If configuredCapacity is already 1, send status 200 to avoid error on client side
+      return { status: 200, message: 'Successfully reset channel capacity to 1' };
+    }
+
+    // If configuredCapacity is already 1, send status 200 to avoid error on client side
+    return { status: 200, message: 'Channel capacity already 1, no adjustment made.' };
   }
 
   return { status: 400, message: 'Invalid adjustment argument' };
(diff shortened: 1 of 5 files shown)


functions/adjustChatCapacity.ts
```diff
@@ -120,6 +121,17 @@ export const adjustChatCapacity = async (
     return { status: 200, message: 'Successfully decreased channel capacity' };
   }
 
+  if (body.adjustment === 'setTo1') {
+    if (channel.configuredCapacity !== 1) {
+      await channel.update({ capacity: channel.configuredCapacity - 1 });
```
murilovmachado, line 126: When I see a condition named `setTo1`, I was expecting the update code to be:
`await channel.update({ capacity: 1 });`

Is there a reason we're doing:
`await channel.update({ capacity: channel.configuredCapacity - 1 });`

functions/adjustChatCapacity.ts
```diff
@@ -32,7 +32,8 @@ type EnvVars = {
 
 export type Body = {
   workerSid?: string;
-  adjustment?: 'increase' | 'decrease' | 'increaseUntilCapacityAvailable';
+  // 'increase' and 'decease' can be removed once backend manual pulling is enabled everywhere
+  adjustment?: 'increase' | 'decrease' | 'increaseUntilCapacityAvailable' | 'setTo1';
```
murilovmachado, line 36: Do we still use the adjustment `decrease` in any scenarios?

murilovmachado (approved):
Looks good to me!
(Just a few questions)

functions/adjustChatCapacity.ts
stephenhand: No, good catch

functions/adjustChatCapacity.ts
stephenhand: Just Flex running with `enable_backend_manual_pulling` set to false. It can be removed once everything is using backend pulls

PR #683, opened 2024-09-05
Fix security hole in transitionAgentParticipants

stephenhand:
## Description

Currently any agent can transition all the agents in a Flex interaction into wrapup, meaning they can potentially end other agent's ongoing tasks. This only affects conversations so does not affect any production helplines

This change only allows the function to be called by workers with the supervisor or who currently hold a reservation for the task

### Checklist
- [X] New tests added
- [N/A] Feature flags / configuration added

### Other Related Issues

None

### Verification steps

Regression test post surveys, regular contact wrapup and chat transfers on conversations
Try to call the transitionAgentParticipant from postman for a task with a token for a worker who is not a supervisor and does not have a reservation for the target task - it should return a 403

### AFTER YOU MERGE


GPaoloni (approved):
LGTM and good catch!

PR #693, opened 2024-09-23
CHI-2961: converation direct transfer queue

stephenhand:
## Description

Use the 'Everyone' queue for direct transfers if it is available to the target worker. This will make the behaviour a little more consistent

### Checklist
- N/A New tests added
- N/A Feature flags / configuration added

### Other Related Issues

None

### Verification steps

Do a direct chat transfer of a conversation but don't accept it yet. 
Check the task on Twilio Console - it should be in the 'Everyone' queue
Repeat for different channels and helplines, it should always be in the 'Everyone' queue if it is visible to the target worker

### AFTER YOU MERGE

### DIFF
--- a/functions/transferChatStart.ts
+++ b/functions/transferChatStart.ts
@@ -270,12 +270,13 @@ export const handler = TokenValidator(
           `Transferring conversations task ${taskSid} to worker ${targetSid} by creating interaction invite.`,
         );
         // Get task queue
+        /*
         const taskQueues = await client.taskrouter
           .workspaces(context.TWILIO_WORKSPACE_SID)
           .taskQueues.list({ workerSid: targetSid });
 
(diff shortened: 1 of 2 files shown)


GPaoloni (commented):
Why is using `Everyone` queue making things more consistent? And what problem is this one trying to solve? :thinking: 

I don't say we shouldn't do this, just bear in mind that the contacts will be saved as "Everyone" in the DB and information about what queue they came from is lost.

stephenhand:
@GPaoloni - sorry, the context for this is that for conversations, direct transfers arbitrarily choose the first valid queue for that worker, so the queue of origin for the task is lost anyway

Because we are using Flex Interactions API rather than the one we were using for programmable chat, we need to set a queue explicitly for direct transfers

GPaoloni (approved):
LGTM

PR #716, opened 2024-11-06
support correct channelTransformations

mythilytm:
### Checklist
- [ ] New tests added

### Other Related Issues

None

### AFTER YOU MERGE


functions/getProfileFlagsForIdentifier.protected.ts
```diff
@@ -96,7 +96,7 @@ const channelTransformations: { [k: string]: TransformIdentifierFunction[] } = {
   modica: [(s) => s.replace('modica:', ''), phoneNumberStandardization],
   facebook: [(s) => s.replace('messenger:', '')],
   messenger: [(s) => s.replace('messenger:', '')],
-  instagram: [],
+  instagram: [(s) => s.replace('instagram:', '')],
   line: [],
   telegram: [],
```
stephenhand, line 101:
```suggestion
  telegram: [(s) => s.replace('telegram:', '')],
```

stephenhand (approved):
From the code it looks like we would need to clip the telegram prefix too. Not so important since I don't think Malta use blocking, sand only they use telegram, but for the future.

PR #725, opened 2024-11-20
fix: handle conversatiosn in sendMessageAndRunJanitor function [CHI-3075]

GPaoloni:
## Description
This PR adds support for Conversations based channel to `sendMessageAndRunJanitor`.

### Checklist
- [x] [Corresponding issue has been opened](https://tech-matters.atlassian.net/browse/CHI-3075)
- [ ] New tests added

### Verification steps
@janorivera tested this and seems like it solves the bug.

### AFTER YOU MERGE

### DIFF
--- a/functions/sendMessageAndRunJanitor.protected.ts
+++ b/functions/sendMessageAndRunJanitor.protected.ts
@@ -54,8 +54,7 @@ export const handler = async (
     if (conversationSid) {
       const conversationWebhooks = await context
         .getTwilioClient()
-        .conversations.services(context.CHAT_SERVICE_SID)
-        .conversations(conversationSid)
+        .conversations.v1.conversations(conversationSid)
         .webhooks.list();
 
       // Remove the studio trigger webhooks to prevent this channel to trigger subsequent Studio flows executions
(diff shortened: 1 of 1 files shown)


functions/sendMessageAndRunJanitor.protected.ts
```diff
@@ -51,28 +51,57 @@ export const handler = async (
...
 
-    const channelWebhooks = await context
-      .getTwilioClient()
-      .chat.services(context.CHAT_SERVICE_SID)
-      .channels(channelSid)
-      .webhooks.list();
+    if (conversationSid) {
+      const conversationWebhooks = await context
+        .getTwilioClient()
+        .conversations.services(context.CHAT_SERVICE_SID)
```
stephenhand, line 57: We just use the default conversations service rather than looking it up via the SID, it has to be the default to be used by flex anyway. Explicitly referencing the API version in the chain helps avoid dependency conflicts in some scenarios Also, explicitly using the `.get` method rather than directly invoking objects as functions makes mocking for unit tests more straightforward.

So this call would be:
```
      const conversationWebhooks = await context
        .getTwilioClient()
        .conversations.v1
        .conversations(conversationSid)
        .webhooks.list();
```

But that's all minor niggles

stephenhand (approved):
LGTM

PR #730, opened 2024-12-12
CHI-3069: Backend contact create

stephenhand:
## Description

* Adds an additional handler to create new contacts when a reservation is accepted, except in the case of offline contact tasks, or tasks created to receive chat transfers
* Loads the form definitions and prepopulates the new contact with data from the task
* Adds the contact ID of the new contact to the task attributes, so the client is aware that the contact has been provided
* All behind an `enable_backend_hrm_contact_creation` feature flag.

We are currently only using the static key encryption on the HRM call. We'll probably need something more secure for production

### Checklist
- [ ] New tests added
- [x] Feature flags / configuration added

### Other Related Issues

None

### Verification steps

Regression test new call & chat contacts, transfers and offline contacts

### AFTER YOU MERGE

### DIFF
--- a/functions/taskrouterListeners/adjustCapacityListener.private.ts
+++ b/functions/taskrouterListeners/adjustCapacityListener.private.ts
@@ -45,41 +45,30 @@ export const shouldHandle = (event: EventFields) => eventTypes.includes(event.Ev
  * @param event
  */
 export const handleEvent = async (context: Context<EnvVars>, event: EventFields) => {
-  try {
-    const {
-      EventType: eventType,
-      WorkerSid: workerSid,
-      TaskChannelUniqueName: taskChannelUniqueName,
-    } = event;
-    if (taskChannelUniqueName !== 'chat') return;
-    console.log(`===== Executing AdjustCapacityListener for event: ${eventType} =====`);
-    const serviceConfig = await context.getTwilioClient().flexApi.configuration.get().fetch();
-    const {
-      feature_flags: {
-        enable_manual_pulling: enabledManualPulling,
-        enable_backend_manual_pulling: enableBackendManualPulling,
-      },
-    } = serviceConfig.attributes;
+  const { WorkerSid: workerSid, TaskChannelUniqueName: taskChannelUniqueName } = event;
+  if (taskChannelUniqueName !== 'chat') return;
+  const serviceConfig = await context.getTwilioClient().flexApi.configuration.get().fetch();
+  const {
+    feature_flags: {
+      enable_manual_pulling: enabledManualPulling,
+      enable_backend_manual_pulling: enableBackendManualPulling,
+    },
+  } = serviceConfig.attributes;
 
-    if (enabledManualPulling && enableBackendManualPulling) {
-      const { path } = Runtime.getFunctions().adjustChatCapacity;
+  if (enabledManualPulling && enableBackendManualPulling) {
+    const { path } = Runtime.getFunctions().adjustChatCapacity;
 
-      // eslint-disable-next-line global-require,import/no-dynamic-require,prefer-destructuring
-      const adjustChatCapacity: AdjustChatCapacityType = require(path).adjustChatCapacity;
-      const body = {
-        workerSid,
-        adjustment: 'setTo1',
-      } as const;
+    // eslint-disable-next-line global-require,import/no-dynamic-require,prefer-destructuring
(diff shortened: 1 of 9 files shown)


functions/taskrouterListeners/adjustCapacityListener.private.ts
```diff
@@ -45,41 +45,30 @@ export const shouldHandle = (event: EventFields) => eventTypes.includes(event.Ev
...
-      const adjustChatCapacity: AdjustChatCapacityType = require(path).adjustChatCapacity;
-      const body = {
-        workerSid,
-        adjustment: 'setTo1',
-      } as const;
+    // eslint-disable-next-line global-require,import/no-dynamic-require,prefer-destructuring
+    const adjustChatCapacity: AdjustChatCapacityType = require(path).adjustChatCapacity;
+    const body = {
+      workerSid,
+      adjustment: 'setTo1',
```
GPaoloni, line 65: Why do we need to set capacity to 1 here? :thinking:

functions/taskrouterListeners/createHrmContactListener.private.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (!eventTypes.includes(eventType)) return false;
+
+  const { isContactlessTask, transferTargetType } = JSON.parse(taskAttributesString ?? '{}');
+
+  if (isContactlessTask) {
+    console.debug(`Task ${taskSid} is a contactless task, contact was already created in Flex.`);
+    return false;
+  }
+
+  if (transferTargetType) {
```
GPaoloni, line 91: Maybe this could reuse the `hasTransferStarted` method (`functions/transfer/helpers.private.ts`)?

functions/taskrouterListeners/createHrmContactListener.private.ts
```diff
@@ -0,0 +1,187 @@
...
+    taskId: taskSid as HrmContact['taskId'],
+    channelSid: channelSid ?? '',
+    serviceSid: (channelSid && serviceConfig.chatServiceInstanceSid) ?? '',
+    // We set createdBy to the workerSid because the contact is 'created' by the worker who accepts the task
+    createdBy: workerSid as HrmContact['createdBy'],
+  };
+
+  const prepopulatePath = Runtime.getFunctions()['hrm/prepopulateForm'].path;
+  const { prepopulateForm } = require(prepopulatePath) as PrepopulateForm;
+  await prepopulateForm(taskAttributes, contactForApi, formDefinitionsVersionUrl);
```
GPaoloni, line 149: I think is harder to follow what is really happening here, cause `prepopulateForm` is mutating the given contact. Could we return a new object to make this explicit? (I will really appreciate that information being evident next time I'm around this code :stuck_out_tongue:)

GPaoloni (approved):
LGTM as far as I can tell

functions/taskrouterListeners/createHrmContactListener.private.ts
stephenhand: Not worth pulling it in for IMHO, when it's a lambda and we can import stuff properly then sure

functions/taskrouterListeners/adjustCapacityListener.private.ts
stephenhand: We always set it back to 1 on certain events. With backend manual pulling, instead of trying to track capacity, we just always have it set to 1, them set it to the value required to take an extra task when manually pulling or accepting transfers, then set it back to 1 again when we're done.

There is no issue with being 'over capacity' other than not being able to accept more tasks, so this is simpler than trying to track capacity against tasks

Also this logic isn't changed in this PR

functions/taskrouterListeners/createHrmContactListener.private.ts
stephenhand: I won't make it a pure function because that makes it significantly uglier for deep objects and there's no practical benefit here since it's a brand new contact anyway.

I thought using 'populate' in the method name would have made the mutation effect clear, bad assumption on my part.

So the function still mutates but we reassign the output to a new variable to be clear. I also changed the method name to be more descriptive

GPaoloni: The reassignment is good enough, I don't mind it being pure as much as being explicit that the contact was transformed. Thanks for the change!

